PR #159, opened 2017-03-04
Adds system.exists

jamonholmgren:
### context.system.exists

Returns 'file' or 'directory' if exists, or null if not

```js
const itExists = context.system.exists('./myfile.wat')
console.log(itExists) // => 'file'
```


skellock:
We have this already on `filesystem.exists`.

https://infinitered.github.io/gluegun/#/context-filesystem?id=exists

PR #179, opened 2017-03-19
Makes spawn asynchronous

### LINKED ISSUE #167: Spawn needs to be async
Spawn can't run with `.sync` because it locks the main thread.  We'll need to spawn async and use callbacks to hook back when we're completed.

abrophy:
Instead of spawning with .sync, spawns asynchronously and uses callbacks to resolve promise.
Fixes #167.


skellock:
Thank you kindly!!

abrophy:
Any time, will remember to lint before future commits

PR #210, opened 2017-09-17
You got semver?  Gimme semver!

GantMan:
Contract semver to the user as officially supported and exposed.


jamonholmgren:
Ooh, I can use this in AM.

PR #212, opened 2017-09-18
don't hate me steve

GantMan:
I think v0.21.1 looks more professional anyway.

### DIFF
--- a/packages/gluegun/test/domain/runtime-extensions.test.js
+++ b/packages/gluegun/test/domain/runtime-extensions.test.js
@@ -6,5 +6,5 @@ test('loads the core extensions in the right order', t => {
   const r = new Runtime()
   const list = pipe(pluck('name'), join(', '))(r.extensions)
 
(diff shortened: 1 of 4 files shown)


packages/gluegun/test/domain/runtime-extensions.test.js
```diff
@@ -6,5 +6,5 @@ test('loads the core extensions in the right order', t => {
   const r = new Runtime()
   const list = pipe(pluck('name'), join(', '))(r.extensions)
 
-  t.is(list, 'strings, print, template, filesystem, system, http, prompt, patching')
+  t.is(list, 'strings, print, template, filesystem, semver, system, http, prompt, patching')
```
GantMan: what exactly is this test doing?

jamonholmgren: Looks like it's checking that `(new Runtime()).extensions` contains those things in that order.

GantMan: It's the order part that got me

PR #215, opened 2017-09-26
Adds defaultCommand to builder and runtime

### LINKED ISSUE #60: Support default commands
Currently a command can't live on it's own.  It needs to live under a namespace.

With `spork`, i made that name space.  And since namespaces aren't commands, you can't just type it.  So I created `default` as the command.

So now it's:

`gluegun spork default`

Which is lame.

I want:

`gluegun spork`.

So we need to build the concept of a default command.  `default.js` would be fine, but I think `index.js` is a little nicer because of the convention on the web and React Native.

jamonholmgren:
This allows for specifying a default command if you don't provide any arguments. This is configured in the runtime builder (rather than specifying a `default.js` like we thought we might do in #60). It made the most sense to me.

image

Fixes #60.


skellock:
Can you add this to the typescript defs and docs too?

PR #216, opened 2017-09-26
Sniffs node version when gluegun is required

### LINKED ISSUE #214: Gluegun should check dependencies upon require
Rather than requiring the CLI author to check Gluegun's dependencies (currently node 7.6+), we should run those sanity checks internally.

Additional sanity checks can be run in the CLI's entry point.

jamonholmgren:
In the interest of making it easier to build a CLI with Gluegun, this PR makes Gluegun handle its own node version sniffing automatically.

Fixes #214 .


skellock:
There are 3 use cases this now eliminates:

* using gluegun as a a library for the static functions it also exports < 7.6
* using it < 7.6 where you polyfill promises and async/await
* using it < 7.6 where you transpile everything via Babel or TypeScript

IMHO, libraries should never call process.exit.

It’s not that I’d say no to this PR, rather, are you really really sure?

GantMan:
Musing on these points:

#### libraries calling process.exit
As a test I typed `node 'HOWDY'` and Node exits with failure, and I feel I expected that.  We'd  rather a CLI be upfront.  I think exiting is fine if a duty cannot be carried out, that the CLI tells me right away.

#### Support for node < 7.6 
I think this is more of a web issue which doesn't fit here.  Anyone with terminal access has a bit more privilege.  With that being said, we should evaluate the trade.

> Knowledge on node requirements from a year ago for zero < 7.6 support 

### TL:DR;
My 2 cents.... Be upfront on no < 7.6 support in docs and outliers will fork, because that's the price of keeping old versions in tech. Technical debt for those who elect it, instead of all.

skellock:
What I'm saying tho is, `gluegun` isn't a CLI.  It's a library.

GantMan:
yeah, I see what you mean.   

As a library to provide a CLI the lines get a bit blurry for me.  My gut pushes me towards simpler solutions in this space because of it.

skellock:
Ya, i seriously doubt anyone will ever use this without a CLI.

The other thing I've also been trying to keep all printing out of the library.  That way the calling CLI gets to customize 100% of the experience.

This would be an exception to that.

Maybe that's not a realistic goal anymore with these changes you guys want to do.

GantMan:
Well, my MO is to think about the 80%+  first.   Because I hate pressing 2 for English when I call a service 😆 

I think that's a good goal, and we can still have it, but maybe there's a special tie-in moment for overriding the defaults.   Good defaults are very helpful, especially for those of us coming to a tool for the first time.

Power users, is usually an effect of a success, and it's awesome!

GantMan:
So if we move the check to the runtime, then a user could bypass or customize the default error!

skellock:
We can't wait that long to check unfortunately.  It needs async/await to even get there.

It needs to be the entry point.

skellock:
https://github.com/typicode/please-upgrade-node  <-- inspiration

jamonholmgren:
> It’s not that I’d say no to this PR, rather, are you really really sure?

So I'd say I'm really really sure, and if someone comes back and files an issue saying "I need < node 7.6 support!" let's change it then.

PR #217, opened 2017-09-26
De-swallow errors

### LINKED ISSUE #107: Add a way to opt-in to exceptions.
Currently we capture errors, tuck them away, and never show them.

https://github.com/infinitered/gluegun/blob/master/packages/gluegun/src/domain/runtime.js#L141

Instead of forcing CLI's to print these with:

```js
const context = await runtime.run()

if (context.error) {
  context.print.error(error.message)
  context.print.debug(context.error)
}
```

Let's add a new builder function called `.rethrowErrorsWhenRunning(true)` that people can opt-into when developing.

jamonholmgren:
~If there's a load error, it displays the error.~
~If there's a runtime error, it tells you there was an error and to pass through `--debug`~
~If there's a runtime error and you pass through `--debug`, it displays the error~
~If there's a runtime error and you pass through `--silent`, it swallows the error (previous behavior)~

We now throw all errors to the plugin author to handle.

Fixes #107. I understand this wasn't your original vision in that issue, @skellock , but this follows the "principle of least surprise" IMO. (High gafometer on this change due to hours of pain.)


packages/gluegun/src/domain/runtime.js
```diff
@@ -159,6 +159,14 @@ async function run (options) {
     try {
       context.result = await context.command.run(context)
     } catch (e) {
+      const { options } = context.parameters
```
skellock: The gluegun runtime should never rely on parameters for its behaviour. It gets its configuration passed down from thee Builder.  That way there is no confusion who owns the command line. It is always the parent app.

packages/gluegun/src/domain/runtime.js
```diff
@@ -159,6 +159,14 @@ async function run (options) {
...
+      if (options.silent || options.quiet) {
+        // shhh...
+      } else if (options.debug || options.verbose) {
+        console.error(e);
+      } else {
+        context.print.error(`There was an error in ${context.plugin.name} ${context.command.name}. Run with --debug to see the error.`)
```
skellock: One of the philospophies of gluegun is that it never prints anything itself.

packages/gluegun/src/loaders/command-loader.js
```diff
@@ -94,6 +94,8 @@ function loadFromFile (file, options = {}) {
       command.errorState = 'badfunction'
     }
   } catch (e) {
+    console.error(e)
```
skellock: I say remove this. But not only this, but the whole feature of not crashing when we have a bad plugin.

skellock:
Thanks for tackling this.  This was the last thing holding up 1.0.

We get this in, and we can launch it.  Then immediately get to work on butchering the interface for 2.0!

jamonholmgren:
@skellock On it.

jamonholmgren:
Whew! That was a lot of work. :P

skellock:
@jamonholmgren so I think we discussed my only objection... we can talk more about it on zoom?  So we good to merge then?  I think we are here.

skellock:
Aww crap wrong fucking pr.

PR #218, opened 2017-09-26
Pass through the name of a plugin when loading it

jamonholmgren:
This prevents `plugin.name` (on the default plugin) from being `src` and lets you do this:

```javascript
  const runtime =
    build()
    .brand('gluegun')
    .loadDefault(`${__dirname}`, { name: 'gluegun' })
    // ...
```


jamonholmgren:
well all right!

Issue #123, opened 2016-12-18
Default commands have unpruned params


GantMan:
> As violent delights have violent ends, so do default commands have unpruned params
> -- Dolores

running `ignite add ui` gives `parameters.first` the value of "default"
![image](https://cloud.githubusercontent.com/assets/997157/21294872/899b3d42-c50d-11e6-997b-ba55f4d73131.png)

This looks like some leakage.

👙 Bikini Bonus:
Would be helpful to have `parameters.last` - which is how I plan on fixing this for futureproof ways.  But I'm not a fan of that method.

skellock:
Ya, I agree, something seems off about parameters.

`parameters` gets rewritten because i'm trying to target the plugin/command.

I'm up for changing this.

skellock:
I think what's bothering me about this is its confusing arguments (things that exist on the command line) with arguments (things that get fed into a command).  I think we can probably clear this up with a conversation.

jamonholmgren:
If we do change this, maybe we should use a different property to stay backwards compatible? Maybe that's a horrible idea but I have `parameters.second` strewn all over my code.

skellock:
This has completely reworked in 2.0

PR #219, opened 2017-09-26
Adds normalizedParams as context.params

jamonholmgren:
Experimental, better params support.

Ref #6 and #123.

Note that this recognizes the plugin name & command name and strips those out (adding them back in as `params.plugin` and `params.command`). It works whether you do `mycli plugin command foo` or `mycli command foo`.

screen shot 2017-09-26 at 12 08 39 am

Discussion welcome, or yolo merging, either way.

### DIFF
--- a/packages/gluegun/src/cli/normalize-params.js
+++ b/packages/gluegun/src/cli/normalize-params.js
@@ -7,7 +7,9 @@ const minimist = require('minimist')
  */
 function normalizeParams (plugin, command, argv = []) {
   // chop it up minimist!
-  const { _: array, ...options } = minimist(argv.slice(2))
+  const options = minimist(argv.slice(2))
+  const array = options._
+  delete options._
 
   // the plugin name or command is sometimes the first/second word -- delete them
   if (array[0] == plugin) { array.shift() }
--- a/packages/gluegun/src/cli/normalize-params.js
+++ b/packages/gluegun/src/cli/normalize-params.js
@@ -1,0 +1,28 @@
+const minimist = require('minimist')
+
+/**
+ * Parses the command-line arguments into a normalized object.
+ *
+ * @return {{}} An object with our awesome keys.
+ */
+function normalizeParams (plugin, command, argv = []) {
+  // chop it up minimist!
+  const { _: array, ...options } = minimist(argv.slice(2))
+
+  // the plugin name or command is sometimes the first/second word -- delete them
+  if (array[0] == plugin) { array.shift() }
+  if (array[0] == command) { array.shift() }
+
+  const first = array[0]
+  const second = array[1]
+  const third = array[2]
+
+  // the rawCommand is the rest of the words
+  const raw = array.join(' ')
+
+  // :shipit:
+  return { argv, first, second, third, raw, array, options }
+}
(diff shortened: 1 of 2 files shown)


jamonholmgren:
If this looks reasonable, I will write some tests and documentation. 👍

packages/gluegun/src/domain/runtime.js
```diff
@@ -141,6 +142,19 @@ async function run (options) {
...
 
+  // normalized params (experimental)
+  const normalizedParams = normalizeParams(
+    context.plugin.name,
+    context.command.name,
+    process.argv
+  )
+
+  context.params = {
+    ...normalizedParams,
```
skellock: Spreading doesn't work on node.

packages/gluegun/src/domain/runtime.js
```diff
@@ -141,6 +142,19 @@ async function run (options) {
...
   context.parameters.string = join(COMMAND_DELIMITER, subArgs)
 
+  // normalized params (experimental)
+  const normalizedParams = normalizeParams(
+    context.plugin.name,
+    context.command.name,
+    process.argv
+  )
+
+  context.params = {
```
skellock: Why not just ditch the old one then?  Having two is confusing.

packages/gluegun/src/domain/runtime.js
jamonholmgren: Aw. :(

jamonholmgren: Wait, it did work, though. Maybe because I have a newer version? 8.5 or something

packages/gluegun/src/domain/runtime.js
jamonholmgren: We can nuke the `parameters` one, but if we replace the name, it will break compat, and I have `context.parameters.second` littered everywhere when it would need to change to `context.params.first` based on the semantics of this.

packages/gluegun/src/domain/runtime.js
jamonholmgren: Replaced with Object.assign. 👍

jamonholmgren:
I'm open to replacing/fixing `parameters`, but I don't want to break Ignite et al, and unless someone wants to volunteer to go fix it on Ignite too ... 😁

skellock:
I think there are already breaking changes.  We don't have to upgrade ignite right away.

I'd rather get the API right now tho.

jamonholmgren:
I'll rename to `parameters`. This means I get to delete more code, @skellock .

skellock:
sounds good!  now is the time!

jamonholmgren:
Major refactor in progress. Closing for now.

packages/gluegun/src/cli/normalize-params.js
```diff
@@ -0,0 +1,30 @@
...
+
+  // the rawCommand is the rest of the words
+  const raw = array.join(' ')
+
+  // :shipit:
+  return { argv, first, second, third, raw, array, options }
+}
+
+module.exports = normalizeParams
+
```
mlaco: 👏

PR #220, opened 2017-09-26
Adds docs and typescript defs for new defaultCommand config

jamonholmgren:
I also added a couple example commands. I will be fleshing those out in future PRs.

Ref #215 .


jamonholmgren:
Reversed in #223 .

PR #221, opened 2017-09-26
Co-locates tests with source files

jamonholmgren:
:tada:


packages/gluegun/src/cli/parse-command-line.test.js
```diff
@@ -1,5 +1,5 @@
 const test = require('ava')
-const parseCommandLine = require('../../src/cli/parse-command-line')
+const parseCommandLine = require('./parse-command-line')
```
jamonholmgren: This is so pretty.

packages/gluegun/src/utils/filesystem-utils.test.js
```diff
@@ -30,7 +30,7 @@ test('isNotDirectory', t => {
 
 test('subdirectories', t => {
   const dirs = subdirectories(`${__dirname}/..`)
-  t.is(6, dirs.length)
+  t.is(dirs.length, 6)
```
jamonholmgren: Ava expects the hard-coded value to be on the right.

skellock: yep.  `actual` and `expected`.

PR #222, opened 2017-09-27
Removes the clipboard.  Too niche.

### LINKED ISSUE #211: Remove clipboard from core
It's way too specific.  My bad.

`gluegun-clipboard` is better.

skellock:
Clipboard doesn't fit the mould for "things a normal cli wants".

Closes #211 .


jamonholmgren:
Would be nice to move this into a plugin at some point.

PR #224, opened 2017-09-28
Nested commands!

jamonholmgren:
* Nested commands are now a thing!
* And they work even if you mix aliases, e.g. `gluegun g model Foo`, with `g` instead of `generate`
* `.test.js` files are excluded from being loaded as a command
* Some other cleanup to make @skellock happy

image


skellock:
Omg! I really wanted this.

Will review when I’m on my computer.

PR #225, opened 2017-09-29
Adds ability for extensions to optionally return a new or modified context

jamonholmgren:
@GantMan and I were discussing how this feels more natural to us as plugin / cli authors -- that we take in a context and return a context.

However, I think allowing mutating the original context and not returning is also a viable strategy, so I'm allowing that as well.

Mutating a context and returning it doesn't really _do_ anything, but it does feel better and seem more logical. context in, mutate, context out.

```js
 /**
  * My extension
  *
  * @param  {RunContext} context The running context.
   */
  module.exports = (context) => {
    context.mything = () => { return 'garbage' }
    return context
  }
```

### DIFF
--- a/packages/gluegun/src/core-extensions/filesystem-extension.js
+++ b/packages/gluegun/src/core-extensions/filesystem-extension.js
@@ -15,6 +15,7 @@ function attach (context) {
   extension.subdirectories = subdirectories
 
   context.filesystem = extension
+  return context
 }
 
 module.exports = attach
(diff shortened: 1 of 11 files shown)


skellock:
So I have a few issues with this.

1. It's fine to have a `void` return.
1. This pattern is common. You'll find this in `koa` and `express`.
1. It's pretending to be a pure function but isn't.
1. You're saying, "don't forget to return the context because it doesn't do anything".

What if this just needs a name change?

Like `beforeCommand` or `beforeRun` or `before`... or `onStart` or something like that?

Would that feel better?

jamonholmgren:
I know, Steve, I know...this just _feels_ better to me.  Don’t think the name change would be better. I also don’t think it will trip people up in practice.

jamonholmgren:
nope

skellock:
HAHAHAHHAHAHHAHhahahaha.

<3

PR #228, opened 2017-09-30
Updates printCommands to be included in context.print

jamonholmgren:
Also works with new command structure and allows refining for nested commands.

image

```js
const { printCommands } = context.print
printCommands(context, ['generate', 'model'])
```

My confidence level that this is the right way to approach this is medium, not high. I think it's _fine_ but there may be a better way. Happy to entertain ideas.

### DIFF
--- a/docs/context-print.md
+++ b/docs/context-print.md
@@ -108,3 +108,19 @@ The text can also be set with the normal printing colors.
 ```js
 spinner.text = context.print.colors.green('i like trees')
 ```
+
+## printCommands
+
+Prints out a table of available commands in a given context.
+
+```js
+const { printCommands } = context.print
+printCommands(context)
+```
+
+You can pass in a "command path" to refine what commands you'd like to see:
+
+```js
+const { printCommands } = context.print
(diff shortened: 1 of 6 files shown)


docs/context-print.md
```diff
@@ -108,3 +108,19 @@ The text can also be set with the normal printing colors.
...
+const { printCommands } = context.print
+printCommands(context)
+```
+
+You can pass in a "command path" to refine what commands you'd like to see:
+
+```js
+const { printCommands } = context.print
+printCommands(context, ['generate', 'model'])
+```
```
jamonholmgren: This has me thinking we need to remove the `.plugins` layer, just use nested commands, like @skellock wants.

Maybe this should be done pre-2.0.0-beta.

jamonholmgren: Because "plugin" could just be the first entry in the `commandPath`.

skellock: Yes!  Tree all the things.

GantMan:
https://i.imgur.com/wqMWK7z.gifv

PR #230, opened 2017-10-01
Adds a stopwatch.

### LINKED ISSUE #197: Add a stopwatch
This is something that's used frequently.

How about `context.system.startTimer()`?  We can [add it here](https://github.com/infinitered/gluegun/blob/master/packages/gluegun/src/core-extensions/system-extension.js#L115).

Usage:

```js
// in your app
module.exports = async context => {
  const timer = context.system.startTimer()
  // ... some more of your slow code
  context.print.info(`finished in ${timer()} ms`)
}
```

I've done something [like this before])(https://github.com/infinitered/reactotron/blob/master/packages/reactotron-core-client/src/stopwatch.js).  Let's harvest that and strip out the `window.performanceNow` browser stuff.

Don't forget [the docs](https://github.com/infinitered/gluegun/blob/master/docs/context-system.md) and tests.

Heads up with the tests though.  If you end up using a `sleep` command to delay for like 100 ms in your test, know that NodeJS is not entirely accurate with it's timers.  They don't guarantee anything +/- 4ms due to how their internal loop dispatches work.

I'd be fine with the test checking +/- 30ms for a 100ms duration.  Sloppy, but honestly, good enough.

skellock:
```js
const finished = context.system.startTimer()

// 1 second later

const elapsed = finished()  // 1000
```

Closes #197

### DIFF
--- a/packages/gluegun/src/core-extensions/system-extension.js
+++ b/packages/gluegun/src/core-extensions/system-extension.js
@@ -94,5 +94,15 @@ module.exports = function (context) {
     return nodeWhich.sync(command)
   }
 
-  context.system = { exec, run, spawn, which }
+  /**
+   * Starts a timer used for measuring durations.
+   * 
+   * @return {function} A function that when called will return the elapsed duration in milliseconds.
+   */
+  function startTimer() {
+    const started = process.uptime()
+    return () => Math.floor((process.uptime() - started) * 1000) // uptime gives us seconds
+  }
+
+  context.system = { exec, run, spawn, which, startTimer }
 }
--- a/packages/gluegun/src/core-extensions/system-extension.test.js
+++ b/packages/gluegun/src/core-extensions/system-extension.test.js
@@ -1,6 +1,8 @@
 const test = require('ava')
 const create = require('./system-extension')
 
+const delay = ms => new Promise((resolve) => setTimeout(resolve, ms))
+
 const context = {}
 create(context)
 const system = context.system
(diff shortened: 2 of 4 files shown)


packages/gluegun/src/core-extensions/system-extension.test.js
```diff
@@ -47,3 +49,15 @@ test('spawn deals exit codes', async t => {
   t.falsy(crap.error)
   t.is(crap.status, 1)
 })
+
+test.serial('start timer returns the number of milliseconds', async t => {
+  const WAIT = 10
+  const NODE_JS_EVENT_LOOP_LAG = 4 // nodejs doesn't guarantee accurate millisecond timing in their event loop
```
skellock: This is fragile. On my computer, (2 year old macbook pro) i didn't see anything under 4ms.

Then I committed this to semaphore and the build failed.  They had 21ms lag.

So I checked the build settings, and we were using `node 7.6` and running the tests in parallel.

I upgraded us to `8.5` (the latest that semaphore supports) and added `-s` to our tests to run serially.

Now it passes.

So perhaps we switch this to be checking only minimum amount of time and not maximum.

It's not knob we can control.

packages/gluegun/src/core-extensions/system-extension.js
```diff
@@ -94,5 +94,15 @@ module.exports = function (context) {
...
 
-  context.system = { exec, run, spawn, which }
+  /**
+   * Starts a timer used for measuring durations.
+   * 
+   * @return {function} A function that when called will return the elapsed duration in milliseconds.
+   */
+  function startTimer() {
+    const started = process.uptime()
+    return () => Math.floor((process.uptime() - started) * 1000) // uptime gives us seconds
```
jamonholmgren: This...this is pretty. @skellock <3 <3 <3

However, did you look at [process.hrtime()](https://nodejs.org/api/process.html#process_process_hrtime_time) or [console.time() / endTime()](https://nodejs.org/api/console.html#console_console_time_label), which may be more suitable / accurate for this type of measurement?

skellock: I used `process.hrtime()` on reactotron.  this is a slightly nicer version of that because we only need a stop & end point.

`console.time()` will have side effects of printing... and actually, now that I think of it, it might not even be available on Node?  I'd have to try that.

skellock: oh, hehe, it is available... you literally linked to it.  ya, looks like the docs say it prints.

jamonholmgren: I just really like this pattern you used here. It's one of the prettiest functions I've ever seen.

PR #232, opened 2017-10-01
Updates lerna and fixes package.json problems

skellock:
A few issues this fixes up:

* Updates `lerna` to the latest (not a beta)
* Ensures `1.0.0` is listed for each dependency.  I've done this in the `1.x` branch already.
* Fixes `lerna bootstrap`. I had `lerna run bootstrap`, which isn't a thing and was causing problems.
* Adds missing `repository` field in `package.json`s


lerna.json
```diff
@@ -1,4 +1,4 @@
 {
-  "lerna": "2.0.0-rc.4",
-  "version": "0.22.0"
+  "lerna": "2.2.0",
```
GantMan: NO MO BETA!

package.json
```diff
@@ -1,19 +1,18 @@
 {
   "name": "gluegun-monorepo",
+  "private": true,
```
GantMan: neat!

package.json
```diff
@@ -1,19 +1,18 @@
...
-  "main": "index.js",
   "scripts": {
     "lint": "lerna run lint",
     "test": "lerna run test",
-    "bootstrap": "lerna run bootstrap",
+    "bootstrap": "lerna bootstrap",
     "publish": "lerna publish"
   },
-  "author": "",
+  "author": "hello@infinite.red",
```
GantMan: Maybe we list our names as contributors?
https://github.com/react-native-community/react-native-linear-gradient/blob/master/package.json#L12-L20

No clue what the benefit is... or if this is worth it.

skellock: imho, this is useless.  Nobody looks at this for attribution. 

How would you feel about implementing this on our projects: https://github.com/kentcdodds/all-contributors  I've been seeing this everywhere now.

package.json
skellock: Ya, i feel dumb for not having this there in the first place.  If someone (probably me) accidentely ran `npm publish` in the root, then this would have been published.  Not a big deal of course.  

It would be a big deal for a private repo tho.  haha.

package.json
GantMan: I like it!

jamonholmgren: I like this too.

PR #233, opened 2017-10-01
Removes max time check as it's too fragile.

skellock:
This test fails on my computer now when i run in parallel mode.

It's not a good candidate for a test unfortunately.  

I removed the top end, but it will still check to see if it's at least the minimum.


skellock:
Oy, ya.. .good call.  Not quite sure why that happened.  I'm using VS Code Insiders build, so I'll double check that i have the `editorconfig` plugin installed.

PR #236, opened 2017-10-04
relative path fix

### LINKED ISSUE #235: Fix relative plugin loads
I'm using gluegun 2.0 beta  and having some issues with relative paths on plugins.

### This will work (absolute path to plugin)
```js
module.exports = async () => {
  build()
    .brand('solidarity')
    .src(__dirname)
    // for testing a plugin
    .plugins('/Users/gantman/Documents/Projects/rn/node_packages/', { matching: 'solidarity-react-*', hidden: true })
    .create()
    .run()
}
```

### This will fail (relative path)
```js
module.exports = async () => {
  build()
    .brand('solidarity')
    .src(__dirname)
    // for testing a plugin
    .plugins('..', { matching: 'solidarity-react-*', hidden: true })
    .create()
    .run()
}
```

This will be a bigger problem when your CLI tries to load its plugins from a common relative path like  the `node_modules` folder:
```js
// Relative PATH O NO!!!!!!
.plugins('./node_modules', { matching: 'solidarity-*', hidden: true })
```

The error I get seems to do fine at finding the plugin... but when it passes it down to be loaded, the relative path is no good.   
image

My suggestion is that all relative paths are converted to absolute paths at when the build identifies a `load` or `loadAll` moment.  That way you don't have to pin down the exact offending spot.

GantMan:
no one reads this.

fixes #235

### DIFF
--- a/packages/gluegun/src/domain/runtime.js
+++ b/packages/gluegun/src/domain/runtime.js
@@ -28,6 +28,7 @@ const { isBlank } = require('../utils/string-utils')
 const { subdirectories, isDirectory } = require('../utils/filesystem-utils')
 const RunContext = require('./run-context')
 const loadPluginFromDirectory = require('../loaders/toml-plugin-loader')
+const { resolve } = require('path')
 
 // core extensions
 const addTemplateExtension = require('../core-extensions/template-extension')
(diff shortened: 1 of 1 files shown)


GantMan:
FYI: @skellock helped find/fix this issue.  But will get no shirt credits from the ocean :(

skellock:
By carefully inspecting where @GantMan's body parts landed, I was able to pinpoint the exact place to apply a patch. 

It was even his idea on how to fix it.

All I did is show up and say, "yup, you should have snipped the blue wire."

PR #239, opened 2017-10-05
No need for argv

GantMan:
Tested in `solidarity` this parameter isn't changing anything.


jamonholmgren:
I would prefer to keep it. :)

GantMan:
even if it does nothing?

jamonholmgren:
It does move where we are retrieving the arguments into the binary, which is where I like it to be. More functional this way.

GantMan:
sounds good

PR #245, opened 2017-10-16
missing type pass along of semver

GantMan:
Merge this one for now, but.....

seems there must be a better way.  Perhaps:
https://github.com/DefinitelyTyped/tsd/blob/master/typings/semver/semver.d.ts


skellock:
Let's using their typings if they're correct.

GantMan:
Sounds good to me... I have no clue how to do that.

GantMan:
Also, kinda strange that we ever tried to compete eh?

PR #246, opened 2017-10-16
Properly stub colors and color types

GantMan:
As you can see here:
https://github.com/infinitered/gluegun/blob/4cf9b6d5067e0bb0fb3619dbb5ef489536966fb2/packages/gluegun/src/core-extensions/print-extension.js#L70

Colors is also color

### DIFF
--- a/packages/gluegun/gluegun.d.ts
+++ b/packages/gluegun/gluegun.d.ts
@@ -588,6 +588,7 @@ export interface GluegunPrintUtils {
    * Colors as seen from colors.js.
    */
   colors: any
+  color: any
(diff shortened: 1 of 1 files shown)


skellock:
That's a shitty api.  Do you want to just deprecate one of them?

GantMan:
I'm not sure why there are two... or how, or what...  But some documentation somewhere got me using `color` over `colors`.  So we'll have to hunt it down.  Might be a good hacktoberfest?

skellock:
I'll merge this in for now until we can get to it.

PR #252, opened 2017-10-21
Reworks normalizeParams to properly handle aliased commands

### LINKED ISSUE #250: Aliases are giving wrong parameters.
I have a command that prints out the `context.parameters`.  I'm seeing different results when i have an alias defined or not.

```shell
am g model
```

Will print these `context.parameters`:

```js
{ plugin: 'app-machine',
  command: 'generate',
  first: 'g',
  second: 'model',
  third: undefined,
  raw: 'g model',
  string: 'g model',
  array: [ 'g', 'model' ],
  options: {},
  argv: [ 'g', 'model' ] }
```

```shell
am generate model
```

Will print these `context.parameters`:

```js
{ plugin: 'app-machine',
  command: 'generate',
  first: 'model',
  second: undefined,
  third: undefined,
  raw: 'generate model',
  string: 'model',
  array: [ 'model' ],
  options: {},
  argv: [ 'generate', 'model' ] }
```

It would be great if `alias` would disappear entirely and be replaced by its full value. The 2nd results appear to be correct.

jamonholmgren:
This is a little smelly and a little dirty in its implementation, but it does work well. If someone wants to take a stab at making it more functional, be my guest. Warning you, though, it's not trivial...

Fixes #250.


packages/gluegun/src/cli/normalize-params.js
```diff
@@ -3,23 +3,18 @@ const minimist = require('minimist')
...
+ * @param {string} plugin     Plugin name
+ * @param {Command} command   The command being run
+ * @param {[]} argv           List of command arguments
+ * @return {{}}               An object with normalized parameters
  */
 function normalizeParams (plugin, command, argv = []) {
   // chop it up minimist!
   const options = minimist(argv)
-  const array = options._
+  const array = command.args
```
jamonholmgren: The command has an artifact attached to it, `.args`, that has what we need. This comes from our command resolver.

🤷‍♂️

packages/gluegun/src/cli/normalize-params.js
```diff
@@ -3,23 +3,18 @@ const minimist = require('minimist')
...
   const raw = array.join(' ')
 
-  // the plugin name or command is sometimes the first/second word -- delete them
-  if (array[0] === plugin) {
-    array.shift()
-  }
-  if (array[0] === command) {
-    array.shift()
-  }
-
```
jamonholmgren: The nice thing is that we don't need to do this wholly inadequate thing anymore.

packages/gluegun/src/domain/runtime-parameters.test.js
```diff
@@ -6,10 +6,28 @@ test('can pass arguments', async t => {
   r.load(`${__dirname}/../fixtures/good-plugins/args`)
   const context = await r.run('args hello steve kellock', { caps: false })
 
-  t.is(context.parameters.string, 'steve kellock')
-  t.deepEqual(context.parameters.command, 'hello')
-  t.deepEqual(context.parameters.plugin, 'args')
-  t.deepEqual(context.parameters.array, ['steve', 'kellock'])
-  t.deepEqual(context.parameters.string, 'steve kellock')
-  t.deepEqual(context.parameters.options, { caps: false })
+  t.is('steve kellock', context.parameters.string)
```
jamonholmgren: The plugin name and command name(s) will not be part of `parameters.string`. If you need that, look into `context.command.commandPath`.

packages/gluegun/src/domain/runtime.js
```diff
@@ -262,14 +262,17 @@ class Runtime {
       commandPath = []
     }
 
+    // track the rest of the commandPath as we traverse
+    const rest = commandPath.slice() // dup
```
jamonholmgren: Not functional. :( Mutating. :( 🤷‍♂️

packages/gluegun/src/domain/runtime.js
```diff
@@ -279,25 +282,30 @@ class Runtime {
         )
       )
       if (cmd) {
+        rest.shift() // remove the current item
```
jamonholmgren: It _is_ simple, though!

packages/gluegun/src/domain/runtime.js
```diff
@@ -279,25 +282,30 @@ class Runtime {
...
+      )
     }
 
-    // looking at the final command path, retrieve the command that matches
-    return find(
-      command => equals(command.commandPath, finalCommandPath),
-      plugin.commands
-    )
+    // attach the rest of the command line arguments to the command
+    targetCommand.args = rest
```
jamonholmgren: Here's where the artifact is attached.

packages/gluegun/src/domain/runtime.js
skellock: Never apologize if it’s self contained within a function.  I don’t care what the internet says.  🥇

packages/gluegun/src/cli/normalize-params.js
skellock: Ya seems a bit odd, I agree.  I’ll add a shrug to this as well.  🤷‍♀️

packages/gluegun/src/domain/runtime-parameters.test.js
skellock: These new assertions are backwards.  They still work of course!  It’s just supposed to be `actual`, `expected` instead of vice versa.

🔁

jamonholmgren: Weird, when I was doing this I had a strange result that I interpreted to mean these were backwards (originally), which is why I flipped them. I'll switch them back. 👍

packages/gluegun/src/domain/runtime.js
jamonholmgren: Totally agree.

jamonholmgren:
So @skellock take a gander at the last commit and see what you think. It's not _quite_ as shoulder-shruggy, but still not super clean.

It's internals, though, probably not a huge deal either way. You make the call, boss.

PR #253, opened 2017-10-22
Negate test files by default

### LINKED ISSUE #247: Unable to put tests in commands folder
Using `gluegun@2.0.0-alpha.2`, I'm unable to keep this file: `commands/tests/list-stacks.test.ts`.  It is trying to load it as a command.

Here's my runtime config.

```ts

const LOAD_OPTIONS: GluegunLoadOptions = { commandFilePattern: '*.{js,ts}' }

/**
 * Sets up the runtime by configuring gluegun.
 */
export function configureRuntime() {
  return build()
    .brand(BRAND)
    .src(resolve(__dirname, '..'), LOAD_OPTIONS)
    .plugin(resolve(`${filesystem.cwd()}`, BRAND), LOAD_OPTIONS)
    .plugins(`~/.${BRAND}/plugins`, LOAD_OPTIONS)
    .create()
}

```

It probably has to do with `LOAD_OPTIONS`.  I'm using `ts-node` and passing in non-standard a `commandFilePattern`.

jamonholmgren:
Fixes #247. You can override this by providing your own `commandFilePattern` or `extensionFilePattern`, which is now recommended to be an array of strings instead of just a string.

### DIFF
--- a/packages/gluegun/gluegun.d.ts
+++ b/packages/gluegun/gluegun.d.ts
@@ -852,21 +852,22 @@ export interface GluegunLoadOptions {
   hidden?: boolean
 
   /**
-   * The file pattern to use when auto-detecting commands. The default is `*.js`.
+   * The file pattern to use when auto-detecting commands. The default is [`*.js`, `!*.test.js`].
+   * The second matcher excludes test files with that pattern.
    *
-   * If you are using `ts-node`, you can switch this to `*.ts` to pick up your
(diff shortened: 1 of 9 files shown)


skellock:
That failing test doesn't appear to be anything related to you.

jamonholmgren:
We could probably include `*.ts` files by default, since that's a pretty common use case.

PR #254, opened 2017-10-22
Adds support for detecting ts files.

skellock:
@jamonholmgren mentioned this in #253.

A `ts-node` environment is still required and not included by default, but if you use the `ts-node` require hook, it's a pretty seamless experience now.

I'm not sure if there's a way to detect `ts-node` and only turn it on then.  That'd be really neat.


skellock:
That fake `xmlhttprequest` failing test thing is weird.

PR #262, opened 2017-10-26
Integrates cli, removes lerna

### LINKED ISSUE #260: Eliminate monorepo, integrate CLI
@skellock and I were talking about the friction involved with coordinating `gluegun-cli` and `gluegun` releases (via lerna), and came to the conclusion that we should probably integrate the `cli` into the `gluegun` package and eliminate the monorepo.

A couple of the issues:

* You currently can't link to the repo in `package.json` dependencies, e.g. `"gluegun": "infinitered/gluegun"
* In [releases](https://github.com/infinitered/gluegun/releases), you can see that there's an alpha.6 release, but it only applies to `gluegun-cli`, not `gluegun` -- confusingly

This should be a relatively straightforward process, and I'll probably tackle it soon.

Input welcome.

jamonholmgren:
Fixes #260. Moves some files around, makes one npm package (we'll deprecate `gluegun-cli`). Tested on several machines under `2.0.0-alpha.13`, even a Windows PC.


package.json
```diff
@@ -1,19 +1,95 @@
 {
-  "name": "gluegun-monorepo",
-  "private": true,
-  "version": "0.0.1",
+  "name": "gluegun",
+  "version": "2.0.0-alpha.13",
```
jamonholmgren: Yes, it took me alphas 7, 8, 9, 10, 11, 12, and 13 to get it right. Don't judge.

jamonholmgren:
Updating semaphore now.

package.json
skellock: Totally judging right now.

PR #265, opened 2017-11-02
adding a method to check for existence in a file

ekryski:
I noticed that most of the methods inside of Ignite have now basically been pushed down to Gluegun. I think this was a good call but there is one that is missing: [`isInFile`](https://github.com/infinitered/ignite/blob/master/src/extensions/patching.js#L80).

I added it to the `patching` extension, made it async, and called it `exists`.

I didn't see an issue for this, are ya'll up for taking something like that?

P.S.

Looks like some of the linting isn't passing. Not a result of my PR here. 😄

### DIFF
--- a/src/core-extensions/patching-extension.js
+++ b/src/core-extensions/patching-extension.js
@@ -7,6 +7,25 @@ const { isFile } = require('../utils/filesystem-utils')
  * @param  {RunContext} context The running context.
  */
 function attach (context) {
+
+  /**
+   * Identifies if something exists in a file
+   *
+   * @param  {string}  filename     The path to the file we'll be scanning.
+   * @param  {string}  findPattern  The case sensitive string or RegExp that identifies existence.
+   * @return {boolean}              Boolean of success that findPattern was in file.
+   */
+  async function exists (filename, findPattern) {
+    let contents = await readFile(filename)
+
+    if (typeof contents !== 'string') {
+      contents = JSON.stringify(contents)
+    }
+
+    let finder = typeof findPattern === 'string' ? new RegExp(`.*${findPattern}.*`, '') : findPattern
+    return !!contents.match(finder)
+  }
+
   /**
    * Updates a text file or json config file. Async.
    *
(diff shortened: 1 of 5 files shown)


ekryski:
Thanks gents and thanks for the quick merge. Appreciate it.

src/core-extensions/patching-extension.js
```diff
@@ -7,6 +7,25 @@ const { isFile } = require('../utils/filesystem-utils')
...
+   * @return {boolean}              Boolean of success that findPattern was in file.
+   */
+  async function exists (filename, findPattern) {
+    let contents = await readFile(filename)
+
+    if (typeof contents !== 'string') {
+      contents = JSON.stringify(contents)
+    }
+
+    let finder = typeof findPattern === 'string' ? new RegExp(`.*${findPattern}.*`, '') : findPattern
```
jamonholmgren: I think I'd prefer to use the built-in `includes` rather than this RegExp:

```js
return typeof findPattern === 'string' ? contents.includes(findPattern) : !!contents.match(findPattern)
```

@skellock @ekryski

skellock: That works too @jamonholmgren.  I can do that if you'd like.  This PR has been merged.

skellock:
If there's questions you have @ekryski, feel free to ask.  We've got a [slack](https://infiniteredcommunity.herokuapp.com/) channel `#gluegun` if github issues doesn't work for ya.

We're still transitioning out of gluegun 1.0 and into 2.0 but @jamonholmgren has done a hell of a job undoing my bad decisions. :rofl: 

@GantMan has built sweet his new [Solidarity](https://github.com/infinitered/solidarity) CLI with 2.x.

I've also got a closed-source project 😿  that's my current active project which uses the `2.x` codebase as well.  This one is generator heavy and uses [Nunjucks](https://github.com/mozilla/nunjucks) rather than the built in `ejs`.

src/core-extensions/patching-extension.js
jamonholmgren: Cool, thanks @skellock -- I would prefer it. Thanks!

jamonholmgren: FYI: the primary reason I don't like the `new RegExp(`.*${findPattern}.*)` way of doing this is if I do:

```js
await context.patching.exists('jamon.jpg')
```

... it will not behave as I would expect, given that the `.` will be treated as a regex special character.

ekryski:
Sounds good. We've got a closed source project using the 2.x Gluegun but are looking at maybe updating the Feathers CLI in the future to use it as well.

src/core-extensions/patching-extension.js
ekryski: @jamonholmgren @skellock Sounds good. I thought that might be an issue as well. We're just using this functionality for pretty basic string matching so I'm happy to put up a quick PR for that change if no one has already done so.

jamonholmgren: Oops, I missed that @skellock [already did this](https://github.com/infinitered/gluegun/pull/267) in #267.

PR #267, opened 2017-11-05
Switches patching.exists to use includes when string

skellock:
@jamonholmgren pointed out that the new `exists` function should check via `includes` when `findPattern` is a string.

Also did a few small uninteresting changes as well:

* switched to `jetpack.read` to load the file instead of `readFile` since don't need `JSON` 
* ^ removed corresponding JSON tests
* added a few early exit sanity checks

### DIFF
--- a/src/core-extensions/patching-extension.js
+++ b/src/core-extensions/patching-extension.js
@@ -20,7 +20,7 @@ function attach (context) {
   async function exists (filename, findPattern) {
     // sanity check the filename
     if (isNotString(filename) || isNotFile(filename)) {
-      false
+      return false
     }
 
     // sanity check the findPattern
--- a/src/core-extensions/patching-extension.js
+++ b/src/core-extensions/patching-extension.js
@@ -1,5 +1,7 @@
 const jetpack = require('fs-jetpack')
-const { isFile } = require('../utils/filesystem-utils')
+const { isFile, isNotFile } = require('../utils/filesystem-utils')
+const { isNotString } = require('../utils/string-utils')
+const { is, test } = require('ramda')
 
 /**
  * Builds the patching feature.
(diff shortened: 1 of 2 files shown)


skellock:
CI is failing because:

```
npm ERR! enoent ENOENT: no such file or directory, rename '/home/runner/gluegun/node_modules/nyc/node_modules/amdefine' -> '/home/runner/gluegun/node_modules/nyc/node_modules/.amdefine.DELETE'
npm ERR! enoent This is related to npm not being able to find a file.
```

I think something is busted over there.  Lemme look.

skellock:
So I fixed CI by adding a new step to the beginning.   

`rm -rf node_modules`


giphy

src/core-extensions/patching-extension.js
```diff
@@ -9,21 +11,35 @@ const { isFile } = require('../utils/filesystem-utils')
...
-   * @return {boolean}              Boolean of success that findPattern was in file.
+   * @param {string} filename The path to the file we'll be scanning.
+   * @param {string} findPattern The case sensitive string or RegExp that identifies existence.
+   * @return {Promise<boolean>} Boolean of success that findPattern was in file.
    */
   async function exists (filename, findPattern) {
-    let contents = await readFile(filename)
+    // sanity check the filename
+    if (isNotString(filename) || isNotFile(filename)) {
+      false
```
ekryski: `return`?

ekryski:
> switched to jetpack.read to load the file instead of readFile since don't need JSON

@skellock It's not the cleanest way to do it but we were hoping to be able to use that functionality to be able to check existence of keys or values in config files.

Alternatively we can just read the file in ourselves and check for existance of keys or values in JS instead of doing string comparison I guess...

src/core-extensions/patching-extension.js
skellock: woah... hahah , ya.. return.  And as it turns out this code effectively does nothing.

skellock:
What was there before (I believe, certainly correct me if I'm wrong!), was a check that would `JSON.stringify()` the contents of the JSON file.  Effectively turning it into a string anyway.

Maybe something like this:

```js
const { pathSatisfies, split, slice, has } = require('ramda')
const { read } = require('fs-jetpack')

export function isKeyInJsonFile (dotPath, filePath) {
  const content = read(filePath, 'json')
  const thePath = split('.', dotPath)
  const upToPath = slice(0, -1, dotPath)
  const lastPath = last(dotPath)

  return pathSatisifies(has(lastPath), upToPath, content || {})
}
```

^ didn't test... might format your computer

PR #269, opened 2017-11-06
Exposing print utils at the top level. Closes #208.

### LINKED ISSUE #208: Missing static spinner export
https://github.com/infinitered/gluegun/blob/master/packages/gluegun/src/index.js#L36

I didn't export the `print` extension like the others, so I'm missing out on the spinner, sadly.

A non-breaking change to repair this would be to export the spinner manually?  Or maybe take a closer look at the interfaces between the `print` extension and the `utils/print`.

It'd be nice to access it like how we would via `context`.

```
import { print } from 'gluegun'
const { spin } = print
```

Oops.

ekryski:
This PR:

- Adds some basic tests for `print-extension`. Not sure how far you wanted to go down that path or not and since Ava was giving me grief I got a bit lazy but at least wanted to test that the interface is defined.

- Adds a test for `print.spin`. Again, not sure how best to test this... deferring to you for advice.
- Exposes `spin` at the top `print` module level and `print-extension` now just uses that.

### DIFF
--- a/gluegun.d.ts
+++ b/gluegun.d.ts
@@ -589,6 +589,10 @@ export interface GluegunPrintUtils {
   newline: () => void
   table: (data: any, options: any) => void
   /**
+   * An `ora` spinner.
+   */
+  spin(options: any): any
+  /**
    * Colors as seen from colors.js.
    */
   colors: any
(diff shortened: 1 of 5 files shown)


skellock:
Considering we had nothing before, I'd say anything is a huge upgrade.

Ya, not sure what's up with coverage.  

I've seen this situation before where not having a source file loaded meant it didn't count towards coverage.   Now that it `print` comes in, it's probably adding that to the formula?  

I'm guessing, I don't know for sure.

This is fine though.

PR #270, opened 2017-11-07
Improves the testing experience

skellock:
Test now run in 4 seconds (down from 30) and don't print anything to the CLI.

Also removed a stray `.only` which was shutting down some tests.

### DIFF
--- a/src/cli/commands/new.js
+++ b/src/cli/commands/new.js
@@ -5,7 +5,6 @@ module.exports = {
   hidden: false,
   run: async (context) => {
     const { parameters, template, filesystem, print, strings, system } = context
-    const { resolve } = filesystem
     const { generate } = template
     const { kebabCase } = strings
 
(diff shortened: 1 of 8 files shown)


src/cli/commands/new.js
```diff
@@ -66,15 +65,15 @@ module.exports = {
...
+      const gen = generate({ template, target, props })
       return prev.concat([ gen ])
     }, active)
 
     // let all generator calls run in parallel
     await Promise.all(active)
 
     // make bin executable
-    filesystem.chmodSync(`${props.name}/bin/${props.name}`, 0755)
+    filesystem.chmodSync(`${props.name}/bin/${props.name}`, '755')
```
jamonholmgren: `+x` _might_ work here, and would be preferable. I didn't test that when I was building it.

jamonholmgren:
> in 4 seconds (down from 30)

[that's fast](https://media1.tenor.com/images/746c4887675c83a2b50a8ec822bdd9a0/tenor.gif?itemid=3563529)

src/cli/commands/new.js
skellock: `standard` didn't like numbers starting with `0`, so i had to change that.  `'755'` works, but ya I hear ya `+x` would be preferrable.

PR #272, opened 2017-11-11
🔨 Reads files sync in patching.exists

### LINKED ISSUE #271: await an async file read with jetpack?
No reason to use async and then await it.

https://github.com/infinitered/gluegun/blob/a9ed51fdffd944ad548e790660ddba5686708e97/src/core-extensions/patching-extension.js#L34

just use `read` which is sync.  Don't add complexity.

skellock:
Also fixes a few standard js issues with patching.

Fixes #271


GantMan:
Looks good to me

GantMan:
Looks good to me too!

GantMan:
Let's merge!

PR #273, opened 2017-11-12
switch out tables

### LINKED ISSUE #207: Switch out ascii-table
Looks like there's a better table lib that would handle colors:

https://github.com/gajus/table

GantMan:
### In this PR
Remove ASCII table and use "Table"

Keep the old table format by default, but allow so much more.

As the name ASCII implies, it sucks at Unicode characters.

Secondly, I wanted to do some advanced table dancing... wait.. wut?

But seriously, I want a Markdown table print option.  So I switched to a configurable table, and configured it.  Added tests for the 3 default modes and kept default behavior the same.

-------
closes #207

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "gluegun",
-  "version": "2.0.0-alpha.13",
+  "version": "2.0.0-alpha.14",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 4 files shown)


GantMan:
Gotcha. What's the upper bound cost for a swap, should I shop other table creators? Or even construct it myself (it doesn't seem hard honestly).

skellock:
There's no official budget size.  We should aim to keep it down though.

The only reason I'm 👎 on this PR is because we used to bring in the whole `lodash` lib, but it's around 4.5mb.  We only use a handful of functions (the ones around string casing), so I switched them to use their [individual packages](https://github.com/infinitered/gluegun/blob/master/package.json#L45-L59) instead.

`table` would undo that.

I agree with you that our current choice is 💩 tho.

I use [`cli-table2`](https://github.com/jamestalmage/cli-table2) on a different project (app-machine).  It works really well and it's about a quarter of the size. It supports the proper string widths even using colours.

It'd be a good fit and it looks like it supports most of the same things as `table`.  Probably wouldn't make your PR look that much different either.

jamonholmgren:
I like the idea of `cli-table2`, and agree with @skellock that we don't want to pull in large dependencies when we don't have to.

skellock:
@GantMan flipped the tables in #280.  Literally.

jamonholmgren:
table flip

PR #283, opened 2017-11-24
Added prettier, husky, lint-staged

jamonholmgren:
I've added prettier and run through the whole code base with it. Also added husky to run `lint-staged` before committing new files. I'll be adding CI linting too, once we get this merged in.


package.json
```diff
@@ -88,8 +85,17 @@
...
-    "node": ">=7.6.x",
-    "npm": ">=4.x"
+  "prettier": {
+    "printWidth": 100,
+    "semi": false,
+    "singleQuote": true,
+    "trailingComma": "es5"
+  },
+  "lint-staged": {
+    "subTaskConcurrency": 1,
```
skellock: You can remove this... if you got it from me, i was struggling one day with `tslint` not formatting "all the file" only "half"... i needed to save several times... i still don't know why that's happening.

jamonholmgren:
@skellock Is it possible to put that ignore rule in the `package.json`? I'm doing some searching but not finding anything.

skellock:
As I understand it, no.  `.prettierignore` settings aren't `prettier` config settings... yet.  i bet there's an issue with this.

skellock:
Also, I think i'm able to put one in my `$HOME` dir so you can say no.

skellock:
I think projects without `prettier` should be forced to put badges on their `readme.md` saying `NO PRETTIER`.  Not vice versa.

PR #286, opened 2017-11-27
🚑 Changes property spread into Object.assign

### LINKED ISSUE #276: Command not found should fail more gracefully
Steps to replicate:

Create a command folder structure like this:

image

Then run `mycli logo`. Fails with an error like this:

```
(node:50164) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): Error: Error: Couldn't load command help -- needs a "run" property with a function.
(node:50164) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

The error is there, but buried in unhandled promise errors.

Here is a good starting point: https://github.com/infinitered/gluegun/blob/master/src/loaders/command-loader.js#L42

skellock:
Since we're targeting 7.6 of node we can't use the `{...thing}` feature.

Fixes #276

### DIFF
--- a/src/utils/print.js
+++ b/src/utils/print.js
@@ -20,12 +20,11 @@ const CLI_TABLE_COMPACT = {
   middle: ' '
 }
 
-const CLI_TABLE_MARKDOWN = {
-  ...CLI_TABLE_COMPACT,
+const CLI_TABLE_MARKDOWN = Object.assign({}, CLI_TABLE_COMPACT, {
   left: '|',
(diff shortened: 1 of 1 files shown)


skellock:
I just switched semaphore over to Node 7.6 so this should catch errors like this in the future.

GantMan:
great catch/figure out!

PR #289, opened 2017-12-12
Inflection Support

ryanlntn:
This PR adds [pluralize](https://github.com/blakeembrey/pluralize) as a dependency and exposes it's functions under string utils. Adding the following functions:

`plural`
`singular`
`addPluralRule`
`addSingularRule`
`addIrregularRule`
`addUncountableRule`
`isPlural`
`isSingular`


ryanlntn:
Yeah sure thing.

skellock:
Sweet!  Thanks!

ryanlntn:
btw @skellock the plural of regex is regexen: https://en.wiktionary.org/wiki/regex

Should I fix?

jamonholmgren:
Excellent. We should update the readme too; I'm happy to do that before releasing.

PR #291, opened 2017-12-13
Properly resolve plugin commands

### LINKED ISSUE #290: Plugin commands aren't resolved properly
When you bring in a plugin, the commands are loaded properly but not resolved when you run that plugin command.

I have a fix for this incoming.

jamonholmgren:
Fixes #290.

Prior to this, the code was doing some strange plugin resolve and then searching for a command in that plugin.

Now, we look through every plugin's commands until we find one that matches, starting with the default plugin (the main CLI).

### DIFF
--- a/src/domain/runtime.js
+++ b/src/domain/runtime.js
@@ -251,7 +251,6 @@ class Runtime {
   /**
    * Find the command for this commandPath.
    *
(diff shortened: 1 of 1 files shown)


jamonholmgren:
I broke something. Working on it...

Issue #288, opened 2017-12-05
Parameters parsing regression


skellock:
I'm having some issues with the parameter parsing in gluegun 2.

With a newly gluegunned project off of `gluegun@2.0.0-alpha.16`, I'm experiencing this:

###  ✅ Command only

```sh
bin/omgparams generate
```

Out of the box, a command by itself is good.

```js
{
  plugin: 'omgparams',
  command: 'generate',
  first: undefined,
  second: undefined,
  third: undefined,
  raw: '',
  string: '',
  array: [],
  options: {},
  argv: ['generate'],
}
```

> HEADS UP: I'm going to remove everything that's `undefined`, `''`, `[]` or `{}` from here on in. It'll be easier to read. They're still there though. Just noise right now.

###  ✅ Alias version

```sh
bin/omgparams g
```

Aliases work as well.

```js
{
  plugin: 'omgparams',
  command: 'generate',
  argv: ['g'],
}
```

###  ✅ Additional parameters

```sh
bin/omgparams generate pizza
```

Even additional parameters work as expected.

```js
{
  plugin: 'omgparams',
  command: 'generate',
  first: 'pizza',
  raw: 'pizza',
  string: 'pizza',
  array: ['pizza'],
  argv: ['generate', 'pizza'],
}
```

### ❌ Options Last

```sh
bin/omgparams generate pizza --now
```

Here's the first problem, when I use options.

The options are leaking into:
 * `second`
 * `string`
 * `array`

(I'm not sure about `raw`.)

```js
{
  plugin: 'omgparams',
  command: 'generate',
  first: 'pizza',
  second: '--now',
  raw: 'pizza --now',
  string: 'pizza --now',
  array: ['pizza', '--now'],
  options: { now: true },
  argv: ['generate', 'pizza', '--now'],
}
```

### ❌ Options First

```sh
bin/omgparams --now generate pizza
```

If i put the --now first, it pushes:

* `generate` into `first`
* `pizza` into `second`

```js
{
  plugin: 'omgparams',
  command: 'generate',
  first: 'generate',
  second: 'pizza',
  raw: 'generate pizza',
  string: 'generate pizza',
  array: ['generate', 'pizza'],
  options: { now: 'generate' },
  argv: ['--now', 'generate', 'pizza'],
}
```

It looks as if `options` are being processed, when they should be lifted out entirely before we start investigating what's left.  This is what `minimist` does for us.

Any thoughts or pointers?

skellock:
[...]

PR #294, opened 2017-12-14
Reworks parameters again, fixes issue #288

jamonholmgren:
This removes a lot of the janky manual parsing of command line arguments that we were doing before and uses `minimist` for all of it.

My confidence level that this works across all situations is not super high, since I did rewrite a fair amount of critical path code here. Tests all pass, and in fact some jankiness with the tests is now looking much more like I'd expect, so that's a good sign.

### DIFF
--- a/src/domain/runtime-parameters.test.js
+++ b/src/domain/runtime-parameters.test.js
@@ -4,7 +4,7 @@ const Runtime = require('./runtime')
 test('can pass arguments', async t => {
   const r = new Runtime()
   r.load(`${__dirname}/../fixtures/good-plugins/args`)
-  const { command, parameters } = await r.run('args hello steve kellock', { caps: false })
+  const { command, parameters } = await r.run('hello steve kellock', { caps: false })
 
   t.is(parameters.string, 'steve kellock')
   t.is(parameters.first, 'steve')
@@ -20,7 +20,7 @@ test('can pass arguments', async t => {
 test('can pass arguments, even with nested alias', async t => {
   const r = new Runtime()
   r.load(`${__dirname}/../fixtures/good-plugins/nested`)
-  const { command, parameters } = await r.run('nested t f jamon holmgren', { chocolate: true })
+  const { command, parameters } = await r.run('t f jamon holmgren', { chocolate: true })
 
   t.is(parameters.string, 'jamon holmgren')
   t.is(parameters.first, 'jamon')
@@ -31,4 +31,18 @@ test('can pass arguments, even with nested alias', async t => {
   t.deepEqual(parameters.array, ['jamon', 'holmgren'])
   t.deepEqual(parameters.options, { chocolate: true })
   t.deepEqual(command.commandPath, ['thing', 'foo'])
+})
+
+test('can pass arguments with mixed options', async t => {
+  const r = new Runtime()
+  r.load(`${__dirname}/../fixtures/good-plugins/args`)
+  const { command, parameters } = await r.run('--chocolate=true --foo -n 1 hello steve kellock')
+  t.deepEqual(command.commandPath, ['hello'])
+  t.is(parameters.string, 'steve kellock')
+  t.is(parameters.first, 'steve')
+  t.is(parameters.second, 'kellock')
+  t.is(parameters.command, 'hello')
+  t.is(parameters.options.foo, true)
+  t.is(parameters.options.n, 1)
+  t.is(parameters.options.chocolate, 'true')
 })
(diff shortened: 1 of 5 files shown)


src/domain/runtime.js
```diff
@@ -60,32 +60,30 @@ async function run (rawCommand, options) {
     commandArray = commandArray.slice(2)
   }
 
-  // find the pluginName and commandName from the cli args
```
jamonholmgren: Removed this whole block of code, using minimist via the `parseParams` function now.

src/domain/runtime.js
```diff
@@ -94,15 +92,6 @@ async function run (rawCommand, options) {
     (this.defaults && this.defaults[context.plugin.name]) || {}
   )
 
-  // normalized parameters
```
jamonholmgren: Not necessary, using the new `createParams` function above.

src/domain/runtime-parameters.test.js
```diff
@@ -4,7 +4,7 @@ const Runtime = require('./runtime')
 test('can pass arguments', async t => {
   const r = new Runtime()
   r.load(`${__dirname}/../fixtures/good-plugins/args`)
-  const { command, parameters } = await r.run('args hello steve kellock', { caps: false })
```
jamonholmgren: I never really liked how we had to put the plugin name here. It's no longer necessary.

I'm not _completely_ sure this is the right direction, though. I'm having some trouble wrapping my mind around the plugins command invocation and what @skellock intended here.

It does feel right, though.

src/domain/runtime-parameters.test.js
```diff
@@ -32,3 +32,17 @@ test('can pass arguments, even with nested alias', async t => {
   t.deepEqual(parameters.options, { chocolate: true })
   t.deepEqual(command.commandPath, ['thing', 'foo'])
 })
+
+test('can pass arguments with mixed options', async t => {
+  const r = new Runtime()
+  r.load(`${__dirname}/../fixtures/good-plugins/args`)
+  const { command, parameters } = await r.run('--chocolate=true --foo -n 1 hello steve kellock')
```
jamonholmgren: New regression test for #288 .

src/domain/runtime-parameters.test.js
skellock: The plugin doesn't make much sense in the new world order.  It's commands all the way down.

src/utils/cli/normalize-params.js
```diff
@@ -1,20 +1,28 @@
 const minimist = require('minimist')
+const { merge } = require('ramda')
+
+function parseParams (argv, extraOpts = {}) {
+  // chop it up minimist!
+  const parsed = minimist(argv)
+  const array = parsed._.slice()
```
skellock: `._.` <--- awww... its so cute

skellock:
One small thing i've noticed.

`am --force generate`

But these do:

```sh
am generate
am generate models
am generate --force
am generate models --force
am generate --force models
```

I don't think it's a deal breaker.  It'd be nice to support this though.  I wouldn't hold up this PR because of it though.

src/domain/runtime-parameters.test.js
jamonholmgren: I’m so glad you said that. It makes my life a lot easier. Ha

src/utils/cli/normalize-params.js
jamonholmgren: Haha didn’t even notice that.

skellock: can't unsee

jamonholmgren:
@skellock So this is a minimist issue, and essentially from what I can tell it's a design decision.

image

See the problem here? It's using the next arg as input. So it'll set `g` to the next argument. minimist wouldn't work with npm either. 😞 🐼 '

Given [this issue](https://github.com/substack/minimist/issues/115), and given that we _already had_ `yargs-parser` as a dependency (through a series of unfortunate events starting with `nyc`), I switched over to `yargs-parser` and removed `minimist`. yargs-parser exhibits the same exact behavior as above, so I think we will lean on the CLI creator to preprocess situations like `npm -g install` before handing it off to gluegun's runtime. Sound reasonable?

jamonholmgren:
Semaphore is having issues, but tests are passing just fine.

skellock:
`yargs-parser` wouldn't have been a production dependency tho because `nyc` is a dev dependency.

But this is all good with me.  `minimist` was internal anyway, so good call.

jamonholmgren:
Oh, you're right. Didn't think about devDependencies. 👍

PR #298, opened 2017-12-16
Adds default version and help commands

### LINKED ISSUE #243: Provide Mount-points for common CLI - Help, Version etc.
In gluegun 2.0 let's handle the common CLI options in a classy way.

## 98% of CLIs
If you want the built-in support, you can just use `commonCommands` function.
image
Maybe even more succinct as `.commonCommands(['help', 'version'])`
These will stop the copy-pasta mess which is mandatory, but not the focus of most CLIs.  You can offload it to us.

## BUT! What if they want to write their own?
**No problem!**, we'll also support `.help()` and `.version()` of every common command, where a user could inject their own function, so they could do something complex like, spit out 2 versions (I'm looking at you `react-native-cli`).  Or simply take control of how things should be displayed.

## No thanks Gluegun, you're getting in my way!!!
**No problem!**, don't use either of the aforementioned functions and just _you do you_.
giphy-downsized

----

Common Commands Should Include `['version', 'help', 'silent']`

### LINKED ISSUE #277: Create a default command that shows help
All we need is a template command with the same name as the cli that load the help command:

```
// src/commands/mycli.js
module.exports = require('./help')
```

jamonholmgren:
Fixes #243, incidentally fixes #277.

I'll go through and annotate this PR, since it's a little larger. The gist of it is that you can do this:

```js
  const cli = build()
    .brand('intern')
    .src(`${__dirname}`)
    .help()
    .version()
    .create()
```

... and that'll add default help and version commands.

image

Not only can you do `.help()`, but you can also provide your own help and version functions if you'd like!

```js
  const cli = build()
    .brand('intern')
    .src(`${__dirname}`)
    .help(context => context.print.info('NO HELP FOR YOU!'))
    .version(context => context.print.info(`99 versions but this is the wrong one, ${context.version()}`))
    .create()
```

Note that we now support `dashed` commands. That is, if you only provide a dashed version of that command, it'll still pick up on it.

```
$ movies --play
```

This would pick up on a `play` command, provided that command has `dashed: true`.


docs/context-api.md
```diff
@@ -1,31 +1,49 @@
...
-module.exports = async function (context) {
-  
-  // great! now what?
-
+module.exports = {
+  name: 'dostuff',
+  alias: 'd',
+  run: async function(context) {
+    // great! now what?
+  },
```
jamonholmgren: The above fixes the example to be true to how we do commands now.

docs/context-api.md
```diff
@@ -1,31 +1,49 @@
...
-**patching**    | manipulating file contents easily                   | fs-jetpack
-**print**       | tools to print output to the command line           | colors, ora
-**prompt**      | tools to acquire extra command line user input      | enquirer
-**semver**      | utilities for working with semantic versioning      | semver
-**strings**     | some string helpers like case conversion, etc.      | lodash & ramda
-**system**      | ability to execute                                  | node-which, execa, cross-spawn
-**template**    | code generation from templates                      | ejs
-
-If this is starting to sound like a scripting language, then good.  That's exactly how to think of it.  Except
-we're not inventing another language.  And we're still running in a `node.js` environment, so you can do whatever you want.
```
jamonholmgren: Updated this documentation to include the new `version` function.

docs/context-print.md
```diff
@@ -116,6 +116,15 @@ The text can also be set with the normal printing colors.
...
 ```
 
+## printHelp
+
+Prints a default help screen, consisting of the brand name, version, and `printCommands` output (next).
+
+```js
+const { printHelp } = context.print
+printHelp(context)
+```
```
jamonholmgren: New command, shows basic help.

docs/plugins.md
```diff
@@ -30,6 +30,7 @@ module.exports = {
   alias: ['a'],
   description: 'Displays the name of an actor',
   hidden: false,
+  dashed: false,
```
jamonholmgren: `dashed` allows us to do `mycli --mycommand`.

gluegun.d.ts
```diff
@@ -17,6 +17,8 @@ export interface GluegunCommand {
   run(context: GluegunRunContext): Promise<any | void>
```
jamonholmgren: I tried, Steve, I tried. This file scares me.

src/cli/commands/help.js
```diff
@@ -1,16 +0,0 @@
-module.exports = {
```
jamonholmgren: We no longer need an explicit `help` command.

src/cli/templates/cli/src/cli.js.ejs
```diff
@@ -10,6 +10,8 @@ async function run (argv) {
     .brand('<%= props.name %>')
     .src(`${__dirname}`)
     .plugins('./node_modules', { matching: '<%= props.name %>-*', hidden: true })
+    .help()     // provides default for help, h, --help, -h
+    .version()  // provides default for version, v, --version, -v
     .create()
```
jamonholmgren: This file is the template CLI for when you run `gluegun new mycli`. Just adding the help/version in here so new CLIs get it for free. Fixes #277.

src/cli/templates/cli/src/commands/help.js.ejs
```diff
@@ -1,12 +0,0 @@
...
-  alias: ['h'],
-  run: async (context) => {
-    const { print } = context
-
-    print.info('')
-    print.info('<%= props.name %> CLI Help')
-    print.info('Coming soon!')
-    print.info('')
-  }
-}
```
jamonholmgren: No longer needed.

src/core-commands/help.js
```diff
@@ -0,0 +1,8 @@
+module.exports = {
+  name: 'help',
+  alias: 'h',
+  dashed: true,
+  run: context => {
+    context.print.printHelp(context)
+  }
+}
```
jamonholmgren: New folder, `core-commands`, provides these core commands (just like `core-extensions`). Thanks @GantMan for the idea.

jamonholmgren: We can probably make this much prettier, btw.

src/core-extensions/core-extension.js
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Gluegun-specific extensions.
+ *
+ * @param  {RunContext} context The running context.
+ */
+function attach (context) {
+  context.version = () => getVersion(context)
+}
+
+module.exports = attach
```
jamonholmgren: This is a new core extension, just called `core-extension.js`, where we can put one-off extensions like `context.version()`.

src/core-extensions/filesystem-extension.js
```diff
@@ -10,8 +10,8 @@ const { subdirectories } = require('../utils/filesystem-utils')
  */
 function attach (context) {
   const extension = jetpack // jetpack
-  extension.eol = os.EOL // end of line marker
-  extension.separator = path.sep // path separator
+  extension.eol = EOL // end of line marker
+  extension.separator = sep // path separator
```
jamonholmgren: Just some code quality improvements here.

src/core-extensions/print-extension.js
```diff
@@ -1,5 +1,5 @@
 const print = require('../utils/print')
-const printCommands = require('../utils/cli/print-commands')
+const { printHelp, printCommands } = require('../utils/print-help')
```
jamonholmgren: Now bringing in `printHelp` and `printCommands` both from `print-help`. I'm doing this whenever I am working on something, removing default exports and adding named exports.

src/domain/builder.js
```diff
@@ -10,6 +10,7 @@ const { loadConfig } = require('../loaders/config-loader')
 class Builder {
   constructor () {
     this.loadPlugins = [] // the plugins
+    this.preloadedCommands = [] // any preloaded commands to add to the default plugin
```
jamonholmgren: This requires a little explanation. So, normally, commands are loaded from files, and we have to go out and require them. But in this case, we have a command object provided to us already (help or version).

... continued below ...

src/domain/builder.js
```diff
@@ -35,11 +36,14 @@ class Builder {
...
-    // the plugins get loaded last
+    // the plugins get loaded next
     this.loadPlugins.forEach(entry => {
       switch (entry.type) {
         case 'default':
-          runtime.loadDefault(entry.value, entry.options)
+          const defaultOptions = Object.assign({}, entry.options, {
+            preloadedCommands: this.preloadedCommands
+          })
+          runtime.loadDefault(entry.value, defaultOptions)
```
jamonholmgren: We pass these preloaded commands through to the _default_ plugin, so they get attached to that plugin.

I thought through a bunch of different ways to do this, but this was the cleanest.

src/domain/builder.js
```diff
@@ -100,6 +104,42 @@ class Builder {
...
+
+  /**
+   * Add a default version handler.
+   * @param  {any} command An optional command function or object
+   * @return {Builder}         self.
+   */
+  version (command) {
+    this.handler('version', command || require(`../core-commands/version`))
+    return this
+  }
```
jamonholmgren: If they provide a command, we use that. Otherwise we reach out to `core-commands` and grab `help` or `version` from there.

src/domain/builder.js
```diff
@@ -100,6 +104,42 @@ class Builder {
...
+      command = {
+        name: type,
+        alias: type[0],
+        run: command,
+        dashed: true // allow running with -v, --version, etc
+      }
+    }
+    this.preloadedCommands.push(command)
+    return this
+  }
```
jamonholmgren: This lets us add these commands and any others we think of later on. Note `dashed: true` here! Also, I'm assuming you want the command alias to be the first letter of the `type` ... could be a lazy decision we undo later, but for now, it's shippable.

src/domain/builder.test.js
```diff
@@ -7,6 +7,12 @@ test('the gauntlet', t => {
     .brand(brand)
     // plugins
     .src(`${__dirname}/../fixtures/good-plugins/threepack`)
+    .help()
+    .version({
+      name: 'gimmedatversion',
+      alias: ['version', 'v'],
+      run: context => 'it works'
+    })
```
jamonholmgren: This is an example of passing in a full command object, rather than just a function.

src/domain/command.js
```diff
@@ -22,6 +23,17 @@ class Command {
...
+  /**
+   * Checks if a given alias matches with this command's aliases, including name.
+   * Can take a list of aliases too and check them all.
+   *
+   * @param {string|string[]} alias
+   */
+  matchesAlias (alias) {
+    const aliases = Array.isArray(alias) ? alias : [alias]
+    return aliases.find(a => this.name === a || this.aliases.includes(a))
+  }
```
jamonholmgren: Added this quality-of-life method to check if a command's name or aliases match anything from a given list of keywords.

src/domain/runtime.js
```diff
@@ -64,7 +65,7 @@ async function run (rawCommand, extraOptions) {
   context.parameters = parseParams(commandArray, extraOptions)
 
   // find the plugin and command, and parse out aliases
-  const { plugin, command, array } = this.findCommand(context.parameters.array)
+  const { plugin, command, array } = this.findCommand(context.parameters)
```
jamonholmgren: `findCommand` can't just look at the parameters `array` anymore. We have to also look at `options` to see any dashed versions.

src/domain/runtime.js
```diff
@@ -179,7 +181,8 @@ class Runtime {
       hidden: options['hidden'],
       name: options['name'],
       commandFilePattern: options['commandFilePattern'],
-      extensionFilePattern: options['extensionFilePattern']
+      extensionFilePattern: options['extensionFilePattern'],
+      preloadedCommands: options['preloadedCommands']
```
jamonholmgren: Have to pass through the preloadedCommands to the `loadPluginFromDirectory` which will attach them to the given plugin.

src/domain/runtime.js
```diff
@@ -244,23 +247,21 @@ class Runtime {
...
+  findCommand ({ array, options }) {
+    const commandPath = array
     let targetPlugin, targetCommand, rest
 
     // start with defaultPlugin, then move on to the others
     const otherPlugins = this.plugins.filter(p => p !== this.defaultPlugin)
     const plugins = [this.defaultPlugin, ...otherPlugins].filter(p => !isNil(p))
 
     targetPlugin = find(plugin => {
-      if (isNil(plugin) || isNilOrEmpty(plugin.commands)) return { command: null, args: [] }
```
jamonholmgren: There were some weird vestigial lines of code in here that I determined were unnecessary. I think. I hope.

src/domain/runtime.js
```diff
@@ -271,8 +272,7 @@ class Runtime {
         const cmd = find(
           command => {
             return (
-              equals(command.commandPath.slice(0, -1), prevPath) &&
-              [command.name].concat(command.aliases).includes(currName)
+              equals(command.commandPath.slice(0, -1), prevPath) && command.matchesAlias(currName)
```
jamonholmgren: Our new `matchesAlias` method in action.

src/domain/runtime.js
```diff
@@ -288,7 +288,15 @@ class Runtime {
...
-        targetCommand = find(command => equals(command.commandPath, [plugin.name]), plugin.commands)
+        // If we're not looking down a command path, look for dashed commands or a default command
+        const dashedOptions = Object.keys(options).filter(k => options[k] === true)
+
+        targetCommand = find(command => {
+          // dashed commands, like --version or -v
+          const dashMatch = command.dashed && command.matchesAlias(dashedOptions)
+          const isDefault = equals(command.commandPath, [plugin.name])
+          return dashMatch || isDefault
+        }, plugin.commands)
```
jamonholmgren: `dashMatch` is the key here. Otherwise, check for a default command like we were before.

jamonholmgren: Note that we always check preloaded commands first. If you don't want a preloaded command to take precedence, don't enable it.

src/index.js
```diff
@@ -33,16 +33,18 @@ require('app-module-path').addPath(process.cwd())
...
-  printCommands
+  printCommands,
+  printHelp
 }
 
 // bring in a few extensions to make available for stand-alone purposes
+require('./core-extensions/core-extension')(context)
 require('./core-extensions/filesystem-extension')(context)
 require('./core-extensions/semver-extension')(context)
 require('./core-extensions/system-extension')(context)
```
jamonholmgren: I feel like this whole section should be removed. It'll take some thought, though. We really only need the `build` function, IMO.

src/loaders/command-loader.js
```diff
@@ -30,7 +30,7 @@ function loadCommandFromFile (file, options = {}) {
   command.name = head(split('.', jetpack.inspect(file).name))
   // strip the extension from the end of the commandPath
   command.commandPath = (options.commandPath || last(file.split('/commands/')).split('/')).map(
-    f => (f === `${command.name}.js` ? command.name : f)
+    f => ([`${command.name}.js`, `${command.name}.ts`].includes(f) ? command.name : f)
```
jamonholmgren: Adds `.ts` again here, gotta support the typescripts

src/loaders/command-loader.js
```diff
@@ -64,4 +64,17 @@ function loadCommandFromFile (file, options = {}) {
...
+  command.name = preload.name
+  command.description = preload.description
+  command.hidden = !!preload.hidden
+  command.alias = preload.alias
+  command.run = preload.run
+  command.file = null
+  command.dashed = !!preload.dashed
+  command.commandPath = preload.commandPath || [preload.name]
+  return command
+}
```
jamonholmgren: New function to make a `Command` object from a hash.

src/loaders/plugin-loader.js
```diff
@@ -44,15 +44,18 @@ function loadPluginFromDirectory (directory, options = {}) {
 
   const jetpackPlugin = jetpack.cwd(plugin.directory)
 
+  // load any default commands passed in
+  plugin.commands = map(loadCommandFromPreload, options.preloadedCommands || [])
```
jamonholmgren: Default/preloaded commands come first.

src/loaders/plugin-loader.js
```diff
@@ -44,15 +44,18 @@ function loadPluginFromDirectory (directory, options = {}) {
...
     const commands = jetpackPlugin
       .cwd('commands')
       .find({ matching: commandFilePattern, recursive: true })
 
-    plugin.commands = map(file => loadCommandFromFile(`${directory}/commands/${file}`), commands)
-  } else {
-    plugin.commands = []
+    plugin.commands = plugin.commands.concat(
+      map(file => loadCommandFromFile(`${directory}/commands/${file}`), commands)
+    )
```
jamonholmgren: Then loaded from files.

src/utils/cli/print-brand-header.js
```diff
@@ -1,26 +0,0 @@
...
-    header(plugin, print)
-    return
-  }
-
-  // otherwise try printing the text file header
-  if (isFile(headerTextFile)) {
-    const text = jetpack.read(headerTextFile)
-    print.fancy(text)
-  }
-}
```
jamonholmgren: We don't use this, and I don't think we should.

gluegun.d.ts
skellock: Haha.  If you think this is scary, you should see libraries like React Native.

docs/plugins.md
skellock: nice!

gluegun.d.ts
skellock: This goes away when we build this in TypeScript.

src/core-extensions/core-extension.js
```diff
@@ -0,0 +1,38 @@
+const jetpack = require('fs-jetpack')
+
+function getVersion (context) {
+  let directory = context.runtime.defaultPlugin && context.runtime.defaultPlugin.directory
```
skellock: Can't wait for the existential operator!  `path` from Ramda is also helpful (although I prefer just doing what you've done here).

src/core-extensions/filesystem-extension.js
skellock: I don't think it's bad quality to bring in a node.js package like that.

src/domain/builder.js
skellock: Ya that's totally fine.  In fact, I'd love to open this up to accept dynamic commands in the future.  I had a use case for this recently!

src/domain/builder.js
skellock: Ya, I'm not a fan of the `alias` is first character.  It's too presumptuous.  It's hard to undo these decisions later, imho.

src/domain/builder.test.js
skellock: Love it.

src/domain/runtime.js
skellock: Ya i find it hard to tell too. This part has been rewritten, so i can't tell you off the top of my head.  

I tried to add a bunch of corner cases in the tests, so hopefully we're good.

src/domain/runtime.js
skellock: Sounds reasonable to me.

src/index.js
skellock: We don't need to expose `core`, but I want to keep the others.

The idea behind the others is sometimes you want to grab some `filesystem` or `string` goodies from a place that doesn't have context.

The idea behind this is a big ol bag of functions.

src/loaders/command-loader.js
skellock: Ha!  I totally hit this last night.  Thanks!

src/loaders/command-loader.js
```diff
@@ -64,4 +64,17 @@ function loadCommandFromFile (file, options = {}) {
   return command
 }
 
-module.exports = { loadCommandFromFile }
+function loadCommandFromPreload (preload) {
+  const command = new Command()
+  command.name = preload.name
+  command.description = preload.description
+  command.hidden = !!preload.hidden
```
skellock: Just pointing out that you didn't use the classic Jamon `Boolean()` here.

src/utils/cli/print-brand-header.js
skellock: Agreed.

src/loaders/command-loader.js
jamonholmgren: Whoops, old habits die hard -- been trying to do `Boolean()` these days!

src/core-extensions/core-extension.js
jamonholmgren: Agreed!

src/loaders/command-loader.js
skellock: Hah! No worries.  I’m trying to switch to `Boolean` as well having heard your opinions.

When I saw you not using them I felt sad and cheated.

src/core-extensions/filesystem-extension.js
jamonholmgren: I don't think it's bad either ... was more thinking along the lines of tree shaking optimizations in the future.

src/domain/builder.js
jamonholmgren: Ooh, interesting. Shouldn't be hard at all. I'll build a command for it in a future PR!

src/domain/builder.js
```diff
@@ -100,6 +104,34 @@ class Builder {
...
+   * @return {Builder}         self.
+   */
+  version (command) {
+    command = command || require(`../core-commands/version`)
+    if (typeof command === 'function') {
+      command = { name: 'version', alias: ['v'], dashed: true, run: command }
+    }
+    this.preloadedCommands.push(command)
+    return this
+  }
```
jamonholmgren: @skellock Reworked this section, removed the `this.handler` method, pass through an explicit alias. I think I like this code better anyway.

src/index.js
jamonholmgren: I'm just not super enthusiastic about building a fake context object every time whether you need it or not. It's also really a different object than the familiar `context` we use elsewhere.

Perhaps we could expose a function that builds this on-demand if you need it rather than every time gluegun runs.

```js
// instead of:
const { filesystem, semver, builder } = require('gluegun')
filesystem.dir('......')

// you'd do this:
const { builder, buildContext } = require('gluegun')
const { filesystem, semver } = buildContext()
filesystem.dir('.....')
```

Just thinking out loud here. I recognize we don't want to change the API just to change the API, but this feels reasonable to me -- accomplish the same thing without building a fake context every time just to throw it away.

src/loaders/command-loader.js
jamonholmgren: It's hard when your role models don't live up to their own ideals.

skellock: 🥛 👃

src/index.js
skellock: We can refactor this without breaking the API.  Make whatever you come up with be an object with those same properties and we're good.

I agree with the smell though.  I did it this way because it was quick (both to implement and execute).  I looks odd how half the object is assembled from the outside, and the other half from the inside.  

Haha.

src/domain/builder.js
skellock: thank you!!

src/index.js
jamonholmgren: If we do an object (rather than a builder function) then we can't load on-demand, right? We'd have to preload the object.

skellock: Correct.  It's super fast though.  Those files will be required anyway, so it'll just hit the cache.

PR #300, opened 2017-12-19
Fixes failed template copy from new

GantMan:
Was playing with master, and failed to create a new project.

I think we have a bigger problem that tests didn't catch it.  Might need to circle back at some point for basic integration tests.


jamonholmgren:
Good call on integration tests. I'll write some with this fix.

PR #301, opened 2017-12-19
Added integration tests and snapshots

jamonholmgren:
Per @GantMan in #300 , this adds an end-to-end integration test. It does slow down the test suite a fair amount (nearly 3x), which is unfortunate, but would have caught the issue in #300 where the current tests do not.

`master` test suite run time: 28.58s
This branch test suite run time: 82.24s

The primary slowdown is because we are running `npm i` on the newly generated CLI.

I think, for now, it's worth the extra test suite time. Open to suggestions.


jamonholmgren:
Still want thoughts, but also want to get a release out tonight.

jamonholmgren:
You had [already written](https://github.com/infinitered/gluegun/commit/716b7f560b7538970f750e3336e77d404b940193#diff-7dc09defaa2e0f46a8ba1fc7b3440847) some mocked-up tests for this part, but they didn't seem to catch this issue, and I haven't looked into why yet.

PR #303, opened 2017-12-19
Adds new meta extension, adds commandInfo & version to it

### LINKED ISSUE #148: Expose the command list as data
Currently printCommands both collects the displays the data.

https://github.com/infinitered/gluegun/blob/master/src/utils/cli/print-commands.js

Which sucks if someone wants to customize that list themselves.

Let's bust these apart and make the data available to DIY printing.

jamonholmgren:
Fixes #148.

I haven't been super happy with the `core-extension` I introduced in an earlier PR, #298. When I thought about it, `version()` and this new `commandInfo()` function are really about the CLI itself. So I decided to create a new extension, `meta`, and move everything there. I'm much happier about it this way.

Note this moves `context.version()` from earlier alphas to `context.meta.version()` so update your code accordingly if you started using it already.

### DIFF
--- a/docs/context-api.md
+++ b/docs/context-api.md
@@ -16,8 +16,8 @@ Here's what's available inside the `context` object you see all over Gluegun.
 
 | name           | provides the...                                    | 3rd party                      |
 | -------------- | -------------------------------------------------- | ------------------------------ |
-| **config**     | configuration options from the app or plugin       |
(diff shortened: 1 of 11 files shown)


skellock:
Does this affect the TS typings too?

jamonholmgren:
Oh shoot, it probably does. Will fix that.

PR #304, opened 2017-12-19
provide default command by default

GantMan:
I think it makes sense to ship them with a default command ready to go.


jamonholmgren:
Interesting, not a bad idea. One problem -- this would be overridden by the `defaultCommand` feature that I just added, and which is already present in the boilerplate `cli.js` file. There are a couple ways to skin this cat: either provide a preconfigured file (like you did here) or provide a `.defaultCommand(<command here>)` in the boilerplate. Or let them decide, and just ship with `.defaultCommand()`.

Let me think about this a bit. I think I still prefer `defaultCommand`, but could be convinced.

jamonholmgren:
I'm going to update the defaultCommand to provide more context-aware feedback.

image

I think I'd rather not ship with a default command in the `commands` folder, but again, I'm a little on the fence about it.

GantMan:
So, the question is.  How many people will want the default command to do something?  I think they all will.

So setting them up to use the file makes sense.  I think I know why you're hessitent - we should discuss.  I think we can get both the 1 file benefit and the starter benefit.

skellock:
I usually hate this sort of thing, but I agree with Gant on this one.

skellock:
caveat:  as long as it can be removed easily... and it can.

GantMan:
> I usually hate this sort of thing, but I agree with Gant on this one.


lol

jamonholmgren:
Okay, you've convinced me. @GantMan Can you fix the semaphore failure and also remove [this line](https://github.com/infinitered/gluegun/blob/master/src/cli/cli.js#L13) please?

GantMan:
Should be good to go, please review.

jamonholmgren:
Thank you very much!

Issue #287, opened 2017-11-30
Add Rollup (or similar) to gluegun-generated CLIs to ship smaller CLIs


jamonholmgren:

jamonholmgren:
I played with it some tonight, but am not finding much info online about making Rollup work with a Node-powered CLI. I was able to get it to run, but it just copied the `src/cli.js` into the build folder and didn't actually roll anything up.

skellock:
Do you have some code I can look at?

jamonholmgren:
I deleted it. No git backup.

![oops](https://media1.tenor.com/images/eabed680d8959e91133d9b286f09415b/tenor.gif?itemid=6118887)

I can pretty easily recreate what I had, though. I'll do that and throw it into a repo for you, @skellock .

skellock:
What I’m thinking is perhaps we start the roll up bundle at the 2nd file and not the cli entry point. 

We have our bin/{brand} file and then our {bundle}.

Not sure if that will matter. 

I’ve used rollup on several libs and it’s awesome. But never on a cli before.

jamonholmgren:
That’s actually what I did, started at the second file (cli.js). It just only included that file, though. Not sure what I’m doing wrong. I’ll post code next week.

jamonholmgren:
@skellock https://github.com/jamonholmgren/gluegun-movies-demo/commit/b071780bd190ce9fb9ac52f5010a2d0088321c03#diff-6e46529c2666be7cf13a3484f3722052R12

jamonholmgren:
Shipped gluegun with Rollup! However, it doesn't close this issue, as we don't include Rollup in generated CLIs. Keeping this zombie caged.

yordis:
@jamonholmgren I would add as part of this improvement the TypeScript definitions to the output.

Even when you are using the JavaScript version, by adding the TypeScript spec some editors could give you an analysis of `gluegun` dependency.

Also, disable `removeComments` from the `tsconfig` since that will remove the inline documentation of the code.

Coming from Elixir, I would love to see npm packages shipping with the inline documentation in the source code. Especially if the package is meant to be used in NodeJS ecosystem where the file sizes do not matter as much as the browser (in any case, your build pipeline could remove the documentation).

I am suffering from this details since either I have to go to Github to figure out the code or I have to put `console.log` and hopefully I can find the info that I need.

jthegedus:
[...]

PR #307, opened 2017-12-22
🏗 TypeScript, Rollup, and Friends! 🔨

### LINKED ISSUE #306: Convert gluegun source to TypeScript
I'm about 80% through [a major refactor](https://github.com/infinitered/gluegun/compare/refactor/internals), which includes converting over to TypeScript. So I'm filing this issue to track some of the items.

1. I'm using `ts-node` right now, and it works pretty well. However, my assumption is that we should keep this as a devDependency and compile a JS bundle ... when other CLIs import `gluegun`, they get the JS bundle. Is that right, or can we assume `ts-node` as a full dependency and call it good?
2. If we do compile a JS bundle, we lose some of the debugging capability in the wild, because we won't get reasonable line numbers.
3. If we do compile, we could conceivably target/support older versions of node. Should we?

jamonholmgren:
# Overview.

This is a major refactor of the internals of Gluegun. The public API _shouldn't_ have changed in this (unless you were doing something really undocumented), but please report bugs.

✅  Added TypeScript
✅  Added Rollup
✅  Updated Folder Structure

Fixes #306 #287 

## Why, Jamon? Why?

I've been working in Gluegun internals for several months now and a few things have started to bug me to the point where I wanted to do a refactor. But I knew if I tried to refactor without some code quality guarantees, I'd cause major issues downstream.

That led me to TypeScript. We've been wanting to make TS a thing on Gluegun for a while now, but it hasn't been much of a priority.

So, I decided to make it a priority. I deleted the gluegun.d.ts file and went to work porting it over to TS. In the process, I discovered we'd need to add Rollup, so that went in as well. Once I got those working, I did my refactors, and here we are.

## TypeScript

I'm using `ts-node` for when you clone down the repo and `yarn link` to get your `gluegun` CLI locally. It works pretty well and lets you test changes quickly.

However, when we go to ship to NPM, that won't work (we don't want to lug along `ts-node` and `typescript` wherever we go. So I include a `yarn compile` script that will compile all the TS files into JS files.

It's not typed in excruciating detail, but the important stuff is mostly there. And it already caught a bunch of issues while I was doing the refactor, including some that were there prior!

## Rollup

Rollup is great! I was able to make it work in a couple hours. Now that I understand it better, I think it'll be easier for me to set it up for other CLIs going forward.

Rollup will take all the files in the freshly-built `build` folder and bundle them into one `dist/index.js`, tree-shaking out old, dead code in the meantime.

_Note that we still have to copy in the `build/cli` folder, since Rollup ignores it, but we need it for gluegun CLI._

Sizes, excluding sourcemap artifacts:

```
528 KB ./src TS source
512 KB ./build compiled JS after tsc
176 KB ./dist bundle after rollup
```

## Updated Folder Structure

I really like the updated folder structure. It's not super drastic, but has some key changes that reduce some of the confusion in my mind. I hope it helps others contribute as well.

```
src
  cli - the gluegun CLI
  core-commands - default commands
  core-extensions - built-in extensions, mostly made up of toolbox tools (below)
  domain - object models used within Gluegun
  fixtures - test fixtures
  loaders - loaders for plugins, commands, extensions, config, etc
  runtime - the Gluegun runtime
  toolbox - a box full of useful tools
  index.js - entry point
```

This isn't super different than before, but the primary difference to me is within `toolbox`. Instead of a loose grab-bag of things we didn't know what to do with, it's now specifically for things that need to be added to `context` via an extension.

The toolbox contains files which end in `-tools.ts`. You can think of them as a drawer in a toolbox, that group related tools. Gluegun then uses those tools directly and also provides them to the host CLI via extensions.

Extensions should be small and basically just load up a related bunch of tools and then stick them on `context`. They shouldn't contain a lot of code (there are still a few which do -- I'll probably refactor those out to the toolbox later).

This really helps me organize the code in my mind.

## What else?

* CI is currently failing and I know why. I'll have to tackle it later, because I may have to rewrite some tests. The tests do pass locally.
* This needs more testing, but tests pass on `solidarity` and app machine. I haven't tried Ignite yet, not sure I want to tackle that beast right now.

## Desired feedback

I know this is a _lot_ of code to review, made worse by some of the formatting style changes. Please at least read through my comments, think about them, and scan the code nearby, looking for anything that stands out to you.

🎩


.all-contributorsrc
```diff
@@ -0,0 +1,129 @@
+{
```
jamonholmgren: This file is used to track contributors by using the [all-contributors](https://github.com/kentcdodds/all-contributors) package by Kent Dodds.

.gitignore
```diff
@@ -7,3 +7,5 @@ yarn.lock
 coverage
 .nyc_output
 generated
+dist/**/*
+build/**/*
```
jamonholmgren: We have two new build artifacts, and we're ignoring them both.

`dist` is where the shipped NPM package and CLI files live.

`build` is where we put JS files created by `tsc` when we compile our TypeScript files.

bin/gluegun
```diff
@@ -1,2 +1,14 @@
 #!/usr/bin/env node
-require(`${__dirname}/../src/cli/cli.js`)(process.argv);
```
jamonholmgren: This is the `gluegun` CLI binary.

* If the `src` folder exists (running locally), we use `ts-node` and the typescript files in `src`
* If not (running via npm install), we use the compiled `dist/index.js` bundle
* We now support `gluegun <whatever> --compiled-gluegun` which allows us to test the compiled bundle rather than using `ts-node`

package.json
```diff
@@ -2,30 +2,41 @@
   "name": "gluegun",
   "version": "2.0.0-alpha.17",
   "description": "Build yourself a pluggable CLI.",
-  "main": "src/index.js",
   "repository": "infinitered/gluegun",
   "bin": {
     "gluegun": "bin/gluegun"
   },
+  "main": "dist/index.js",
```
jamonholmgren: `main` now points to the `dist` folder, rather than the `src` folder (which is _not_ packaged up into the NPM package at all).

package.json
```diff
@@ -2,30 +2,41 @@
...
-    "snapupdate": "ava --update-snapshots",
-    "coverage": "nyc --reporter=lcov --reporter=html --reporter=text ava",
-    "coveralls": "nyc ava && nyc report --reporter=text-lcov | coveralls",
-    "format": "prettier --write \"**/*.js\" && standard --fix",
-    "start": "node src/index.js",
-    "lint": "standard",
-    "shipit": "np --yolo"
+    "format": "prettier --write \"**/*.ts\" && tslint -p . --fix",
+    "compile": "tsc -p .",
+    "rollup": "rollup -c && cp -a ./build/cli ./dist/",
```
jamonholmgren: `yarn rollup` not only bundles up the javascript files, but also copies over the `cli` files as-is, because we can't bundle those up if we want to use them.

package.json
```diff
@@ -2,30 +2,41 @@
...
+    "build": "yarn format && yarn compile && cp -a ./src/cli/templates ./build/cli/ && yarn rollup",
```
jamonholmgren: This command gets used a lot -- basically run this anytime you want to update your `dist` package.

package.json
```diff
@@ -2,30 +2,41 @@
...
     "name": "Infinite Red, Inc.",
     "email": "npm@infinite.red",
     "url": "https://github.com/infinitered/gluegun"
   },
-  "files": ["docs", "bin", "src", "sniff.js", "LICENSE", "readme.md", "gluegun.d.ts"],
-  "types": "gluegun.d.ts",
+  "files": [
+    "docs",
+    "bin",
+    "dist",
```
jamonholmgren: Note we are _not_ including `src` anymore, but are including `dist`.

package.json
```diff
@@ -59,44 +70,67 @@
...
+    "rollup-plugin-json": "^2.3.0",
+    "rollup-plugin-node-resolve": "^3.0.0",
+    "rollup-plugin-progress": "^0.4.0",
+    "rollup-plugin-re": "^1.0.6",
+    "rollup-plugin-uglify-es": "^0.0.1",
     "sinon": "^4.0.1",
-    "standard": "^10.0.2",
-    "temp-write": "^3.3.0"
+    "temp-write": "^3.3.0",
+    "ts-node": "^4.0.2",
```
jamonholmgren: `ts-node` is a devDependency, because we ship JavaScript, not TypeScript!

package.json
```diff
@@ -59,44 +70,67 @@
...
-  "standard": {
-    "parser": "babel-eslint"
-  },
   "prettier": {
-    "printWidth": 100,
+    "printWidth": 120,
     "semi": false,
     "singleQuote": true,
-    "trailingComma": "es5"
+    "trailingComma": "all"
```
jamonholmgren: With formatting...just tell me what to do and I'll do it. This seemed to work the best for me.

rollup.config.js
```diff
@@ -0,0 +1,123 @@
+import resolve from 'rollup-plugin-node-resolve'
```
jamonholmgren: Rollup is awesome! The build step basically goes like this:

* `tslint` with `prettier` and `standard` rules formats the code base
* `tsc` compiles 1-to-1 all `src/**/*.ts` files into their corresponding `build/**/*.js` files
* `rollup` goes in and bundles all files from `build` into one `dist/index.js` file, tree-shaking to remove dead code
* We also bring in `build/cli/*` into `dist/cli/*` as-is because rollup ignores that folder

This file is quite well commented, so I'll let you read the comments.

scripts/string-docs.js
```diff
@@ -2,7 +2,7 @@
  * A script that helps generate some content for the strings documentation.
  */
 
-const strings = require('../packages/gluegun/src/utils/string-utils')
+const strings = require('../packages/gluegun/src/toolbox/string-tools')
```
jamonholmgren: `utils` is now `toolbox`. Read the main pull request text body to find out why.

sniff-async.js
```diff
@@ -0,0 +1,6 @@
+/**
+ * When we require this file, it'll blow up if we don't support async/await
+ */
+export default async function() {
+  await new Promise(resolve => resolve())
+}
```
jamonholmgren: I had to move this file to the root to avoid it getting bundled up and de-`async`'d in the process.

sniff.js
```diff
@@ -1,20 +1,24 @@
 // check the node version
 const semver = require('semver')
 
-// the structure to return
```
jamonholmgren: Minor changes to this file to make it survive the gauntlet.

src/cli/cli.test.ts
```diff
@@ -1,6 +1,6 @@
-const test = require('ava')
-const cli = require('./cli')
-const sinon = require('sinon')
+import test from 'ava'
```
jamonholmgren: A lot of `const x = require('x')` were converted to `import x from 'x'`. Not all. I don't entirely understand _why_ `import` didn't work always, but it didn't, and thus you will see some leftover `require`s. I'm happy to receive an education if you want to enlighten me why sometimes `import` doesn't work with some of these.

src/cli/templates/cli/src/cli.js.ejs
```diff
@@ -8,11 +8,10 @@ async function run (argv) {
   const cli =
     build()
     .brand('<%= props.name %>')
-    .src(`${__dirname}`)
+    .src(__dirname)
```
jamonholmgren: You win Steve.

src/cli/templates/cli/src/cli.js.ejs
```diff
@@ -22,4 +21,4 @@ async function run (argv) {
   return context
 }
 
-module.exports = run
+module.exports = { run }
```
jamonholmgren: Again, my lack of knowledge about imports and exports and requires shows here. I had to use `module.exports =` in several cases because it just didn't work otherwise. I'd get `undefined` or some load error.

src/core-extensions/print-extension.js
```diff
@@ -1,73 +0,0 @@
-const print = require('../utils/print')
```
jamonholmgren: This isn't a delete, but rather a rename with enough changes that git got confused.

While you're looking at this, notice it's a 73 line file. I removed a lot of the duplication and made it a ... (see below)

src/core-extensions/print-extension.ts
```diff
@@ -0,0 +1,13 @@
...
+ * Extensions to print to the console.
+ *
+ * @param  {RunContext} context The running context.
+ */
+function attach (context) {
+  // attach the feature set
+  context.print = print
+}
+
+module.exports = attach
```
jamonholmgren: ... a 13 line file. Much nicer. This is due to the `toolbox`, which I talk about in the main post.

src/cli/templates/cli/src/cli.js.ejs
jamonholmgren: I have some theories I might pursue at a later date. I think dynamic requires aren't supported by `import` so it causes issues if you try to dynamically require a file using import.

src/core-extensions/template-extension.test.ts
```diff
@@ -1,10 +1,10 @@
...
-const Runtime = require('../domain/runtime')
-const { startsWith } = require('ramdasauce')
+import test from 'ava'
+import { startsWith } from 'ramdasauce'
+import Runtime from '../runtime/runtime'
 
 const createRuntime = () => {
   const r = new Runtime()
-  r.load(`${__dirname}/../fixtures/good-plugins/generate`)
+  r.addPlugin(`${__dirname}/../fixtures/good-plugins/generate`)
```
jamonholmgren: I renamed a few internal methods, like `load` is now `addPlugin`. Just clearer to me.

src/core-extensions/template-extension.ts
```diff
@@ -55,10 +56,10 @@ function attach (context) {
     const content = ejs.render(templateContent, data)
 
     // save it to the file system
-    if (!stringUtils.isBlank(target)) {
+    if (!stringTools.isBlank(target)) {
       // prep the destination directory
       const dir = replace(/$(\/)*/g, '', target)
-      const dest = `${jetpack.cwd()}/${dir}`
+      const dest = jetpack.path(dir)
```
jamonholmgren: This was a bugfix -- the previous version caused some issues.

src/domain/builder.js
```diff
@@ -1,164 +0,0 @@
-const Runtime = require('./runtime')
```
jamonholmgren: This file wasn't deleted, just renamed with lots of changes so git got confused.

src/domain/builder.ts
```diff
@@ -0,0 +1,125 @@
+import Runtime from '../runtime/runtime'
```
jamonholmgren: The new Builder is more pass-through. Instead of gathering and staging a bunch of configuration and then using `create` to create a runtime and configure it, it generally just dispatches the configuration over to its internal `runtime` and is much simpler.

I wasn't able to entirely drop it due to one factor: `build().brand('my-brand')`. On the `runtime`, we have code relying on `runtime.brand` being a property, not a setter function.

I have a solution. We could stop doing `context.runtime.brand` in our CLIs and move to `context.meta.brand` (using the new meta #303 extension). Then we can repurpose `runtime.brand` into a setter function that returns the `runtime`, and we're good to delete the `Builder` entirely and have the fluent API operate directly on a new `RunTime` object without too much API breakage.

However, I don't mind this file too much, now. It doesn't do a lot, and what it does is useful. So I'm okay leaving it as-is. Let me know what you think.

src/domain/extension.ts
```diff
@@ -1,13 +1,18 @@
...
+
 /**
  * An extension will add functionality to the context that each command will receive.
  */
-class Extension {
+export default class Extension {
+  public name?: string
+  public description?: string
+  public file?: string
+  public setup?: (context: RunContext) => void
```
jamonholmgren: You can see here some TypeScript bindings. I didn't use the previous gluegun typings, for the most part. Partially because I wanted to use TypeScript and learn (the hard way, as is my standard operating procedure) why we'd want to type everything. As it was, I ended up doing a lot of TypeScript typings and really enjoyed the process.

I almost certainly did things wrong here. I'm still a TS noob. Please feel free to tell me where I went wrong. However, it compiles, runs, and catches a lot of silly mistakes already!

src/domain/options.ts
```diff
@@ -0,0 +1,6 @@
+/**
+ * A flexible object for the many "options" objects we throw around in gluegun.
+ */
+export default interface Options {
+  [key: string]: any
+}
```
jamonholmgren: This is that `Flexible` (here just called `Options`) interface I was talking to Steve about in Slack.

src/domain/plugin.js
```diff
@@ -1,19 +0,0 @@
-/**
```
jamonholmgren: Renamed, not deleted. Thanks git!

src/domain/run-context.ts
```diff
@@ -0,0 +1,70 @@
+import { GluegunRunContext } from '../index'
```
jamonholmgren: Nothing too amazing here, except that my typings are probably making Steve cringe due to their looseness. That was a design goal, btw. To make Steve cringe.

src/domain/runtime.js
```diff
@@ -1,315 +0,0 @@
-const { parseParams, createParams } = require('../utils/cli/normalize-params')
```
jamonholmgren: The Runtime and its associated files and tests were moved into `src/runtime/*`. It's important enough I want it in its own folder.

src/fixtures/good-modules/async-function.js
```diff
@@ -4,7 +4,7 @@ const after = time => {
   })
 }
 
-async function hi () {
+async function hi() {
```
jamonholmgren: You can skip most of the fixtures. Nothing much to see here.

src/index.ts
```diff
@@ -0,0 +1,89 @@
...
+//
+// https://github.com/patrick-steele-idem/app-module-path-node
+// https://gist.github.com/branneman/8048520
+//
+// This adds the node_modules for `gluegun` to the "search path".
+//
+// So, wherever folks have their scripts (plugins or their apps), they will be
+// given a chance to resolve dependencies back up to gluegun's node_modules.
+//
+// I may be going to hell for this, but I'm taking you bastards with me.
```
jamonholmgren: Still makes me laugh.

src/index.ts
```diff
@@ -0,0 +1,89 @@
...
+  printCommands,
+  printHelp,
+  filesystem,
+  semver,
+  system,
+  prompt,
+  http,
+  template,
+  patching,
+}
```
jamonholmgren: I had to export things like this in order for TypeScript to believe me that these were available.

GantMan:
OMG. Please don’t ever review like this again.  lol 

My phone vibrated on every note you made.

src/loaders/home-plugin-directories.js
```diff
@@ -1,26 +0,0 @@
-const { map } = require('ramda')
```
jamonholmgren: This file isn't used anywhere. Deep-filed.

src/loaders/module-loader.js
```diff
@@ -1,14 +0,0 @@
-const { isBlank } = require('../utils/string-utils')
-const { isNotFile } = require('../utils/filesystem-utils')
-const throwWhen = require('../utils/throw-when')
-
-// try loading this module
-function loadModule (path) {
-  throwWhen('path is required', isBlank, path)
-  throwWhen(`${path} is not a file`, isNotFile, path)
```
jamonholmgren: I removed the very cute `throwWhen` function and replaced with a couple `if` statements, which you can see a couple files below. That function was only used here, and I didn't think it warranted its own file and import.

src/loaders/module-loader.ts
```diff
@@ -0,0 +1,17 @@
...
+import { isBlank } from '../toolbox/string-tools'
+
+// try loading this module
+function loadModule (path) {
+  if (isBlank(path)) {
+    throw new Error('path is required')
+  }
+  if (isNotFile(path)) {
+    throw new Error(`${path} is not a file`)
+  }
```
jamonholmgren: Here are the two `if` statements instead of `throwWhen`. More understandable to inexperienced coders, I think.

GantMan:
Jamon you have to be stopped!

src/loaders/project-plugin-directories.js
```diff
@@ -1,15 +0,0 @@
-const { subdirectories } = require('../utils/filesystem-utils')
```
jamonholmgren: Deep-filed because unused.

src/runtime/run.ts
```diff
@@ -0,0 +1,68 @@
+import { clone, isNil, merge } from 'ramda'
```
jamonholmgren: I moved this function from its former place in the `runtime.js` file into `run.ts`. It seemed important enough to warrant its own place.

It's refactored to be easier to understand. Also sprinkled in some comments.

src/runtime/runtime-find-command.ts
```diff
@@ -0,0 +1,81 @@
+import { equals, find, isNil, reduce, sort } from 'ramda'
```
jamonholmgren: I moved this function out of `runtime.js` into its own file, as well. It's a really important and complex function. I tried commenting enough to keep it maintainable, but I'm afraid it's still one of the more opaque pieces of code in gluegun. Luckily, it mostly "just works" at this point, so we shouldn't have to touch it much going forward. 🤞🏻

GantMan:
You can’t be stopped.

src/runtime/runtime.ts
```diff
@@ -0,0 +1,174 @@
+import Command from '../domain/command'
```
jamonholmgren: Here's the updated `Runtime` with `run` and `findCommand` removed.

src/runtime/runtime.ts
```diff
@@ -0,0 +1,174 @@
...
+    this.addCoreExtensions()
+  }
+
+  /**
+   * For backwards compatability. No-op.
+   * @returns {Runtime} This runtime.
+   */
+  public create (): Runtime {
+    return this
+  }
```
jamonholmgren: This was added _in case_ we end up wanting to remove the `Builder`.

src/runtime/runtime.ts
jamonholmgren: The methods are all pretty reasonably-sized now, by the way.

src/runtime/runtime.ts
```diff
@@ -0,0 +1,174 @@
...
+
+    // find matching subdirectories
+    const subdirs = subdirectories(directory, false, options.matching, true)
+
+    // load each one using `this.plugin`
+    return subdirs.map(dir => this.addPlugin(dir, dissoc('matching', options)))
+  }
+}
+
+export default Runtime
```
jamonholmgren: It should be pretty readable code now.

GantMan:
In the year 2049 post apocalyptic nuclear fallout.  There are roaches, styrofoam, and Jamon bot 90000

src/toolbox/meta-tools.ts
```diff
@@ -0,0 +1,82 @@
+import * as jetpack from 'fs-jetpack'
```
jamonholmgren: The toolbox is where we put the real fun of `gluegun`. We attach it all in `extensions` and fire it up with the `Runtime`, but this is the good stuff.

All files are named `./src/toolbox/*-tools.ts` and you can think of them as drawers in a toolbox, full of tools you can pick out when you want them and bundle them together to pass out.

GantMan:
I’m turning off my phone for the night.  You win.

src/utils/async-await-check.js
```diff
@@ -1,7 +0,0 @@
-/**
```
jamonholmgren: Moved to root dir as `sniff-async.js`.

src/utils/cli/normalize-params.js
```diff
@@ -1,45 +0,0 @@
-const yargsParse = require('yargs-parser')
```
jamonholmgren: Moved to `toolbox/parameter-tools.ts`.

src/utils/cli/print-banner.js
```diff
@@ -1,31 +0,0 @@
-const print = require('../utils/print')
```
jamonholmgren: Deleted because we don't use this.

src/utils/command-info.js
```diff
@@ -1,52 +0,0 @@
-const { pipe, map, sortBy, prop, propEq, reject, replace, unnest, equals } = require('ramda')
```
jamonholmgren: Moved into `toolbox/meta-tools.ts`

src/utils/get-version.js
```diff
@@ -1,34 +0,0 @@
-const jetpack = require('fs-jetpack')
```
jamonholmgren: Moved into `toolbox/meta-tools.ts`.

src/utils/print-help.js
```diff
@@ -1,32 +0,0 @@
-const print = require('./print')
```
jamonholmgren: Moved into `toolbox/print-tools.ts`

tsconfig.json
```diff
@@ -0,0 +1,29 @@
+{
```
jamonholmgren: TypeScript config file

tsconfig.json
```diff
@@ -0,0 +1,29 @@
...
+  "include": [
+    "src/**/*.ts"
+  ],
+  "exclude": [
+    "node_modules",
+    "src/fixtures",
+    "*.test.ts",
+    "*.md"
+  ]
+}
```
jamonholmgren: Every line earned its place in sweat, tears, and blood.

tslint.json
```diff
@@ -0,0 +1,12 @@
...
+  "rules": {
+    "trailing-comma": [true],
+    "no-string-throw": false,
+    "strict-type-predicates": false,
+    "no-console": false,
+    "no-var-requires": false,
+    "object-literal-sort-keys": false,
+    "interface-name": false
+  }
+}
```
jamonholmgren: I don't know what to do here, but these rules were what finally did the trick for me.

jamonholmgren:
Whoops...I had no idea this was sending out an email every time I commented! Sorry about that, the 13 watchers on this repo. I will use Github's "Review" batching next time, my bad.

make it stop

skellock:
### Self Reviews

I'd prefer to read the design decisions in the body as opposed into the implementation details.  It makes the thread hard to follow.  If you use code blocks and/or permalinks in the body, it reads more like a story.

Another problem that this causes is important design decisions can get buried if a further commit changes a line that's being discussed.  Github will sort of archive that discussion.

### Big Ball Of Mud

Ya, you're right this is too big to digest.  

You did a refactor + formatting pass + language change + build tooling.  That's 4 PRs in one.  More if you consider smaller things like all contribs.  

### Tests

I would love to see the tests pass.  There looks to be 2 different sets of failures.  Ones for directory counts and ones for colored terminal output.

For the colors, you could use [strip-ansi](https://www.npmjs.com/package/strip-ansi) in your test for simplicity.  

For the filesystem failures, not quite sure what's going on there.  I've had problems with this in the past when I've changed my .gitignore rules though.  Not sure if a fresh clean is needed on semaphore or not.  Might want to look at that if nothing pops out as wrong to you in code.

### Linting Rules

I'm not a fan of `tslint-config-standard`.  They bring too much to the table I don't agree with.  For example, the `interface-name` rule is so weird.  Even the TypeScript team recommends not doing that.

For linting, I like to know what I'm getting into... otherwise you're left with installing random rules to opt-out and you don't know exactly why (e.g. the ones you had to put in).  

That's your call tho!

### Rollup

I'm not sure we need both `uglify` and `google-closure-compiler`.  `uglify` is much lighter dependency.  I believe closure compiler uses (or used to back in the day at least) java.  

Downside of `uglify` is they have several times release broken packages causing the internet to crumble. If we can pin it, that'd be awesome.

It looks like you might not be using either tho in the rollup config, so this is more FYI if you decide on using them.

### Regarding Imports

Why sometimes `import * as Lol from 'lol'` instead of `import Lol from 'lol'`?  

There's a big backstory behind this.  So here's the quick version:

Babel made a decision on how to handle some ways to import because TC39 didn't.  By the time TC39 did ratified the spec, they decided on a way incompatible with Babel's way.  TypeScript switched over to the "official" way but introduced a `allowSyntheticDefaults` config option to play nice. Babel said, "lolumadbro?"

This is reason number 700 on my list of: seriously, just used named exports always.

(btw: take my explanation with a understanding that this not the whole story.  there's more to it than that.  Team babel have every right to push the language forward and sometimes these kinds of breaks are inevitable.)


### TL;DR

Bottom line is, I trust you, so I'm going to approve this.

package.json
```diff
@@ -59,44 +70,67 @@
...
+    "tslint-config-standard": "^7.0.0",
+    "typescript": "^2.6.2"
   },
   "ava": {},
-  "standard": {
-    "parser": "babel-eslint"
-  },
   "prettier": {
-    "printWidth": 100,
+    "printWidth": 120,
```
GantMan: I know default is 80, and Steve bumped to 100.... would like to hear thoughts on why this went to 120, is it better?

src/cli/cli.test.ts
GantMan: import default is actually named under the covers as "default", so I believe you can do `import { default as Nachos } from 'x'` I hear default imports are the devil... so *shrug* we'll all be learning as we go.

src/cli/templates/cli/src/cli.js.ejs
GantMan: I approve of this change

src/core-extensions/patching-extension.ts
```diff
@@ -87,11 +87,11 @@ function attach (context) {
    * Convenience function for replacing a string in a given file. Async.
    *
    * @param {string} filename       File to be prepended to
-   * @param {string} replace        String to replace
+   * @param {string} strReplace     String to replace
```
GantMan: hungarian notation here?  Seems a bit off-key.  You mean `stringToReplace` so maybe make the var names like this:

`filename, oldContent, newContent`
or
`filename, findContent, replaceContent`

just a nit

src/core-extensions/print-extension.js
GantMan: to fix this rename, you can do `git mv`, which will force git to understand the rename, then do the changes.

src/core-extensions/template-extension.test.ts
GantMan: 💖

src/core-extensions/template-extension.ts
GantMan: I might have been fighting this exact issue in solidarity, which made me bring in callsite.  I'm a bit out of context now, but I would have to identify plugin paths with callsite in order to fin their templates.

src/domain/run-context.ts
GantMan: cringe

src/loaders/module-loader.ts
GantMan: you can still make these cute(ish)
```js
if (isBlank(path)) throw new Error('path is required')
if (isNotFile(path)) throw new Error(`${path} is not a file`)
```

src/toolbox/print-tools.test.js.md
```diff
@@ -0,0 +1,75 @@
...
+      [
+        '[32msuccess!![39m',
+        undefined,
+      ],
+      [
+        '[31merror...[39m',
+        undefined,
+      ],
+      [
+        '[31mv[39m[33mv[39m[32mv[39m [35m-[39m[31m-[39m[33m-[39m[32m-[39m[34m-[39m[35m[[39m [33mD[39m[32mE[39m[34mB[39m[35mU[39m[31mG[39m [32m][39m[34m-[39m[35m-[39m[31m-[39m[33m-[39m[32m-[39m [35mv[39m[31mv[39m[33mv[39m',
```
GantMan: AVA should strip out terminal colors with ansi-strip... they annoy me when they do this, and it causes snapshot failures :/

GantMan: Maybe we should swap to jest?
https://shift.infinite.red/switching-from-ava-to-jest-for-typescript-a6dac7d1712f

src/utils/cli/print-banner.js
GantMan: reminds me of this ugly crap:  https://github.com/infinitered/gluegun/pull/35

lol, glad we didn't use it.

tsconfig.json
GantMan: you can add comments in here, you should.  It helps explain how each line earned its place.

tsfmt.json
```diff
@@ -0,0 +1,4 @@
+{
+  "indentSize": 2,
+  "tabSize": 2
+}
```
GantMan: wut is this file?  plz don't tell me another format file is needed.

package.json
```diff
@@ -59,44 +70,67 @@
...
     "prettier": "^1.9.2",
+    "rollup": "^0.52.3",
+    "rollup-plugin-closure-compiler-js": "^1.0.6",
+    "rollup-plugin-commonjs": "^8.2.6",
+    "rollup-plugin-filesize": "^1.5.0",
+    "rollup-plugin-json": "^2.3.0",
+    "rollup-plugin-node-resolve": "^3.0.0",
+    "rollup-plugin-progress": "^0.4.0",
+    "rollup-plugin-re": "^1.0.6",
+    "rollup-plugin-uglify-es": "^0.0.1",
```
GantMan: did you consider https://github.com/rollup/rollup-plugin-typescript ????

rollup.config.js
```diff
@@ -0,0 +1,123 @@
...
+  // we also output a sourcemap, because we want to make sure we can debug reasonably well
+  output: {
+    file: 'dist/index.js',
+    format: 'cjs', // commonjs for node
+    sourcemap: true,
+  },
+
+  // rollup complains if we don't list out all the external dependencies
+  // here you go, rollup. now shut up
+  external: externalPackages(),
```
GantMan: EEEEEEK!  NO LIKEY THIS....

src/cli/templates/cli/src/cli.js.ejs
ryanlntn: @jamonholmgren That's correct. There is a proposal to include dynamic imports now though and it is possible to setup in babel: https://babeljs.io/docs/plugins/syntax-dynamic-import/

Of course you can also just use `require`.

src/core-extensions/print-extension.js
jamonholmgren: Yeah, that's what I originally did. Then I squashed it so you all wouldn't get on my case about 18 "WIP" commits, and that was all lost.

src/loaders/module-loader.ts
jamonholmgren: @GantMan Yeah, I tried. Prettier slapped my face and changed it to the above. 🤷‍♂️

tsconfig.json
jamonholmgren: In a json file? I thought comments weren't allowed.

tsfmt.json
jamonholmgren: I pulled this from App Machine, will probably delete it.

bin/gluegun
```diff
@@ -1,2 +1,14 @@
...
+
+// check if we're running in dev mode
+var devMode = require('fs').existsSync(`${__dirname}/../src`)
+var wantsCompiled = process.argv.indexOf('--compiled-gluegun') >= 0
+
+if (devMode && !wantsCompiled) {
+  // hook into ts-node so we can run typescript on the fly
+  require('ts-node').register({ project: `${__dirname}/../tsconfig.json` })
```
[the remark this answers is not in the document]
jamonholmgren: @GantMan Correct, ts-node basically builds in-memory as you run it, so no distinct build step is needed. It's fine for testing things but of course we use `tsc` and `rollup` to create the real distributed files.

tsconfig.json
GantMan: tsconfig uses JSON5 [http://json5.org/](http://json5.org/)

You can do all kinds of reasonable things, but only in this file.  Other JSON files will not get such awesome treatment.

package.json
jamonholmgren: Only if you add more question marks

jamonholmgren: Just FYI, it doesn't quite work for what I need from it, as I need to compile the CLI files without rolling them up.

src/toolbox/print-tools.test.js.md
jamonholmgren: I'll strip ansi colors. I kind of wanted to keep them in, as we're literally testing ansi colors. But ... yeah.

src/core-extensions/patching-extension.ts
jamonholmgren: Linting complained that `replace` was already defined, so I was just trying to use a different var name.

package.json
GantMan: # ?
❤️

package.json
jamonholmgren: Completely arbitrary. Tell me what to do and I'll do it.

tsconfig.json
jamonholmgren: Unfortunately, I forgot as soon as I committed the file. I'll leave it for a future exercise.

jamonholmgren:
@skellock 

✅ Self Reviews - gotcha, will do this a different way going forward
✅ Big Ball Of Mud - agreed, but not really sorry, this was totally worth it
✅ Tests - passing
✅ Rollup - updated, removed extraneous rollup plugins
✅ Regarding Imports - fixed in latest commit
✅ TL;DR - ❤️
❓ Linting Rules - I really just need to know what to use -- what should I use, if not `tslint-config-standard`? I was going off of Solidarity, but I'm happy to go with AM rules if that's what you recommend.

jamonholmgren:
@ryanlntn I am too, just have additional rules with `tslint-config-standard` right now.

GantMan:
@jamonholmgren - I feel prettier and standard have different domains that interact.  It's probably best to keep our standar-js ways AND prettier.

jamonholmgren:
In this particular case (as discussed in chat earlier) they're competing directly. We should probably just go with one or the other.

Again, 0 gafo, it's just standing in my way for merge. 🤷‍♂️

skellock:
I don’t like `tslint-config-standard` for several reasons. I mention that above. There are some rules that conflict with prettier, some rules that aren’t from standard at all, some rules that are arbitrary, and some that I just don’t agree with.  

That library also has a dependency on another set of rules. Then we’re taking those two sets we don’t control and adding another set of exclusions on top.

`tslint-config-prettier` is only about subtracting rules that clash with prettier.

I loved `standard` for js.  I feel like it’s not a strong fit for `typescript` since `tsc` does so much for us. But, my approach for linting in `ts` has been opt-in rather than opt-out.  That’s backwards from my opinion in `js` because of the amount of foot guns in vanilla `js`.

I say, let’s start fresh. Give prettier first dibs. Add a handful of tslint rules that doesn’t mess with prettier and then start adding more 1 by 1 deliberately when we encounter real gotchas.

With `lint-staged` in place, this is pretty much moot anyway.

Go with your gut.

PR #308, opened 2018-01-04
yarn build creates types from TS

GantMan:
I consumed gluegun 2x beta1 and though it worked as a package, my reference to the exposed types was corrupted.  This brings back types by automatic generation.

image

PR #309, opened 2018-01-05
Bundles up ts typings with dts-bundle

jamonholmgren:
Per discussion in #308, this bundles up the typescript types in one file using [dts-bundle](https://github.com/TypeStrong/dts-bundle). Not sure how good the types are, but it does produce convincing output.


skellock:
The `.d.ts` files that `tsc` will have relative `import` statements.  That's how they connect all the files together.  Your `types` in `package.json` will point to the top of that tree (often coming out of an `index.ts` file).

I wouldn't sweat chasing a bundled `.d.ts`.  These are dev dependencies.

scripts/bundle-d-ts.js
```diff
@@ -0,0 +1,7 @@
+var dts = require('dts-bundle')
+
+dts.bundle({
+	name: 'gluegun',
```
GantMan: dat space dough

GantMan:
I'm agreeing with Steve.  Unless you can show a significant benefit, I say drop the d.ts bundle.  It's cool that it exists, but I highly doubt you're getting your technical debt's worth.

jamonholmgren:
No problem. This is apparently an area [with a fair amount of disagreement](https://github.com/Microsoft/TypeScript/issues/17196#issuecomment-315397965) over in TypeScript land, so I'll go with the de facto standard here with lots of files. (I like the cleanliness of 1 file, but agree in practice it won't have much of an impact).

Issue #312, opened 2018-01-08
Simpler Named Exports


skellock:
In JavaScript it was kinda easier to [wrap up exports in an object](https://github.com/infinitered/gluegun/blob/0e368e24fdc0a80d0098dc1a6bf5750db8893d88/src/toolbox/string-tools.ts#L70-L99), however in TypeScript, I find it often simpler to just do it inline.  It will compile down to the right thing.

For example:

```
function pascalCase(value) {
  return pipe(camelCase, upperFirst)(value)
}
```

Would change to:

```
export function pascalCase(value) {
  return pipe(camelCase, upperFirst)(value)
}
```

And then you could drop the [whole bottom half](https://github.com/infinitered/gluegun/blob/0e368e24fdc0a80d0098dc1a6bf5750db8893d88/src/toolbox/string-tools.ts#L70-L99) of that file.

jamonholmgren:
I couldn't figure out how to export [this part](https://github.com/infinitered/gluegun/blob/0e368e24fdc0a80d0098dc1a6bf5750db8893d88/src/toolbox/string-tools.ts#L19-L28) of the file. Any suggestions?

jamonholmgren:
Fixed in #319 and #324 .

Issue #311, opened 2018-01-08
Re-exporting


skellock:
Instead of [importing just to export](https://github.com/infinitered/gluegun/blob/0e368e24fdc0a80d0098dc1a6bf5750db8893d88/src/toolbox/string-tools.ts#L1-L17), you can do that in one step by changing the first word from `import` to `export`.

```ts
export * as camelCase from 'lodash.camelcase'
```

jamonholmgren:
Nice! I think my problem was I tried to do `export cameCase from 'lodash.camelcase'` which didn't work. womp womp

jamonholmgren:
So `export * as whut from 'whut'` isn't supported right now, although there are a [couple](https://tc39.github.io/proposal-export-ns-from/) [proposals](https://tc39.github.io/proposal-export-default-from/) currently hanging out in stage 1.

I *may* be able to simplify the syntax somewhat though.

skellock:
I must have gotten it confused with `export * from 'whut'`.  😭 

I'm sad.

PR #319, opened 2018-01-09
Replaces old requires with imports where possible

### LINKED ISSUE #310: Prefer `import` over `require`
I recommend sticking with `import` rather than [`require`](https://github.com/infinitered/gluegun/blob/0e368e24fdc0a80d0098dc1a6bf5750db8893d88/src/index.ts#L24-L30) as much as possible to keep the types.

`require` does have it's place for sure though.  Such as, guarding with an `if` statement or intentionally dodging bad types in a 3rd party lib.  Unlike `require`, `import` gets hoisted so it's not as flexible, but checks that huge static analysis checkbox!

jamonholmgren:
This PR replaces a few `require` statements with `import` instead, and exports in a more friendly way. Fixes #310, #311, #312.

## Remaining requires

I have a few remaining requires. These are either in test fixtures (whatevs -- because node doesn't like `import` yet) or requiring dynamic files, like extensions or commands.

## Core extensions and commands

Note that the core extensions and commands are `import`ed. This is because they are known ahead-of-time.

*Question: will this be confusing to people looking at our source code?* They have to write their extensions and commands as commonJS modules, but we write them as es6 modules.

Example:

```js
// our core-commands/new.ts
export default {
  run: (...) => {
    // ...
  },
}

// their commands/new.js
module.exports = {
  run: (...) => {
    // ...
  }
}
```

I _can_ keep our internal commands and extensions as commonJS modules. No real preference. Input requested.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,16 +1,3 @@
-// --- Hackasaurus Rex --------------------------------------------------------
-//
-// https://github.com/patrick-steele-idem/app-module-path-node
-// https://gist.github.com/branneman/8048520
-//
-// This adds the node_modules for `gluegun` to the "search path".
-//
-// So, wherever folks have their scripts (plugins or their apps), they will be
-// given a chance to resolve dependencies back up to gluegun's node_modules.
-//
-// I may be going to hell for this, but I'm taking you bastards with me.
-// ----------------------------------------------------------------------------
-
 // first, do a sniff test to ensure our dependencies are met
 const sniff = require('../sniff')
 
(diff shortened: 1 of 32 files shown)


src/index.ts
```diff
@@ -1,16 +1,3 @@
-// --- Hackasaurus Rex --------------------------------------------------------
-//
-// https://github.com/patrick-steele-idem/app-module-path-node
-// https://gist.github.com/branneman/8048520
```
skellock: 😿

src/toolbox/string-tools.ts
```diff
@@ -1,39 +1,55 @@
...
+import * as lodashSnakeCase from 'lodash.snakecase'
+import * as lodashStartCase from 'lodash.startcase'
+import * as lodashTrim from 'lodash.trim'
+import * as lodashTrimEnd from 'lodash.trimend'
+import * as lodashTrimStart from 'lodash.trimstart'
+import * as lodashUpperCase from 'lodash.uppercase'
+import * as lodashUpperFirst from 'lodash.upperfirst'
+import * as pluralizeModule from 'pluralize'
+
+export const camelCase = lodashCamelCase
```
skellock: Does `export * as camelCase from 'lodash.camelcase'` work?  1 step instead of 2.

skellock:
Re: cjs vs es modules... 

Yes it's confusing.  But only because it's already complicated.  

`commonjs` is on the way out and TS is saying to us: "do it the new way and we'll compile down to the right module system if you tell me to".

It's beautiful.

src/toolbox/string-tools.ts
jamonholmgren: It does not, I fought it but it kept telling me this:

image

src/index.ts
```diff
@@ -62,28 +48,13 @@ attachHttpExtension(context)
...
+export const filesystem = context.filesystem
+export const semver = context.semver
+export const system = context.system
+export const prompt = context.prompt
+export const http = context.http
+export const template = context.template
+export const patching = context.patching
+export const print = context.print
+export const generate = context.generate
+export const strings = context.strings
```
ryanlntn: It's too bad you can't use the spread operator here. `export { ...context }` would be nice.

jamonholmgren: Yeah, it really would.

PR #324, opened 2018-01-11
Simplifies exports in print and string tools

### LINKED ISSUE #311: Re-exporting

jamonholmgren:
Fixes #311 (as much as we can right now, at least).

### DIFF
--- a/src/toolbox/string-tools.ts
+++ b/src/toolbox/string-tools.ts
@@ -1,39 +1,39 @@
 import { is, isEmpty, pipe } from 'ramda'
 
-import * as lodashCamelCase from 'lodash.camelcase'
-import * as lodashKebabCase from 'lodash.kebabcase'
-import * as lodashLowerCase from 'lodash.lowercase'
-import * as lodashLowerFirst from 'lodash.lowerfirst'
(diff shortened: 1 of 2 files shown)


src/toolbox/string-tools.ts
```diff
@@ -1,39 +1,39 @@
...
+export { padEnd }
+export { padStart }
+export { repeat }
+export { snakeCase }
+export { startCase }
+export { trim }
+export { trimEnd }
+export { trimStart }
+export { upperCase }
+export { upperFirst }
```
jamonholmgren: Hm, perhaps I should just `export { cameCase, kebabCase, ... }` instead.

jamonholmgren:
> Totally overthinking it. I'm merging this before you keep going.

😆 I totally was.

PR #330, opened 2018-01-15
Fix table print bug

GantMan:
Looks like a line of code found it's way in.

No tests provided this round.  We should really do a full pass at some point.

## Before:
image

## After:
image


GantMan:
Good news is this tripped the snapshot test!  Bad news is, you must have broken this and said "i'm ok with that" to update the snapshot, lol.

Fixing snapshot now.

jamonholmgren:
Hm, I don't think this is entirely correct. We need to look at this section of code closer.

I had some issues with typing the original line. Specifically, the part that shows this:

`[tableArray.options.head, ...tableArray]`

image

Because of that, my assumption was I needed to do this:

```js
[tableArray.options.head, ...tableArray]

// became

[tableArray.options.head].concat(tableArray)
```

Perhaps the `...tableArray` wasn't doing anything before? This part of the code is really confusing to me, and just deleting the `.concat()` without knowing why won't quite cut it.

jamonholmgren:
I also don't remember updating the snapshot at any point during this process.

GantMan:
image

PR #339, opened 2018-01-30
Adds an optional context types to the Command

orta:
Warning funky types ahead. Will comment inline. Fixes #339

### DIFF
--- a/src/domain/command.ts
+++ b/src/domain/command.ts
@@ -1,19 +1,25 @@
 import { RunContext } from './run-context'
 
-export interface GluegunCommand {
+export interface GluegunCommand<TContext extends RunContext = RunContext> {
+  /** The name of your command */
   name?: string
+  /** A tweet-sized summary of your command */
   description?: string
-  run: (context: RunContext) => void
+  /** The function for running your command, can be async */
+  run: (context: TContext) => void
+  /** Should your command be shown in the listings  */
   hidden?: boolean
   commandPath?: string[]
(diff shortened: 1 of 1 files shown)


src/domain/command.ts
```diff
@@ -1,19 +1,25 @@
 import { RunContext } from './run-context'
 
-export interface GluegunCommand {
+export interface GluegunCommand<TContext extends RunContext = RunContext> {
```
orta: `TContext` is a generic type, which should extend `RunContext` and if you don't include it the `<XXYY>` after your GluegunCommand, you get `RunContext`

src/domain/command.ts
```diff
@@ -1,19 +1,25 @@
...
+  /** The name of your command */
   name?: string
+  /** A tweet-sized summary of your command */
   description?: string
-  run: (context: RunContext) => void
+  /** The function for running your command, can be async */
+  run: (context: TContext) => void
+  /** Should your command be shown in the listings  */
   hidden?: boolean
   commandPath?: string[]
```
orta: I couldn't find docs internally about this, so someone else will have to doc it

src/domain/command.ts
jamonholmgren: This is interesting, I didn't know this was possible in TS. Very cool.

src/domain/command.ts
jamonholmgren: I can doc it!

PR #347, opened 2018-02-06
 👮‍♀️ Simplifies commands to a flat list rather than nested in plugins

### LINKED ISSUE #344: Adding defaultCommand prevents non-core plugins from working
### Problem

Using `gluegun@2.0.0-beta.4`, if you add register a `.defaultCommand()` on your builder, it will prevent from running commands in other plugins.

### Scenario

After adding a `.plugin()`, the `Runtime` shows the 2 plugins.  The first being the default set, the 2nd one being the one I add.

If you try to run it with `mycli plugin2 validcmd` the `.defaultCommand()` runs instead.

You can see it fine, you just can't run it.

### Workaround

Don't use `.defaultCommand()` (unfortunately).

### Where This May Be Happening

I think the problem is in `findCommand` within `src/runtime/runtime-find-command.ts`.

Here's how I think it's going down.

* The [default plugin](https://github.com/infinitered/gluegun/blob/eeb40321305ace72f214a15ab2fe12234dad5a17/src/runtime/runtime-find-command.ts#L25)  is taking precendent.
* [finalCommandPath](https://github.com/infinitered/gluegun/blob/eeb40321305ace72f214a15ab2fe12234dad5a17/src/runtime/runtime-find-command.ts#L39) evals to `[]` because this command doesn't match anything
* We then hit the if statement happy path and [wind up fishing up the default command](https://github.com/infinitered/gluegun/blob/eeb40321305ace72f214a15ab2fe12234dad5a17/src/runtime/runtime-find-command.ts#L66)
* Then we [bail](https://github.com/infinitered/gluegun/blob/eeb40321305ace72f214a15ab2fe12234dad5a17/src/runtime/runtime-find-command.ts#L75) on our `find` because we found it... preventing the loop into our 2nd plugin (where the command really lives).

### Random Thoughts

I wonder if we could do a 2 pass search.  Identify the plugin first, and then search for the command within?

jamonholmgren:
Commands are now kept in a flat list in the Runtime. This is a step in the right direction, although plugins aren't removed. They're just a bit less important now.

This should fix #344.

Only a net gain of 6 lines of code!

image


skellock:
I just wanted to explain myself a little better about the whole "smelly" comment... I meant "tightly coupled", not "you suck"...

Here's a possible way out if you want to explore this:

## The Goal

User types a `string` and we match it to a `Command`.

But `Command`s can live in `Plugin`s and can have any amount of `depth`.

This structure could work.

```ts
interface CommandMap {
  [location: string]: Command
} 
```

Break up your current `findCommand()` into two parts:  `build()` and `find()`.

## Step 1: Build Map

`build()` would be about creating a list of all known possible options.

Let's say we're building a "hypothetical" CLI called `am`.  It ships with `.src()` commands:

#### Commands

- `generate`
- `help`
- `lint`

Our `CommandMap` at this point might look like this:

```ts
{
  'am generate': Command,
  'am help': Command,
  'am lint': Command,
}
```

#### Aliases

A couple of them have aliases though, so it looks more like this:

```ts
{
  ...whatItWasBefore,

  'am g': Command,
  'am h': Command,
}
```

#### Nested Commands

But then, some jerk named @ryanlntn adds some nested commands under `stories` which has 2 commands (with 1 having an alias).

```ts
{
  ...whatItWasBefore,

  'am stories generate': Command,
  'am stories g': Command,
  'am stories reset': Command,
}
```

#### Default Plugin

But remember we used `.src()` to add this, so this is the default plugin.  That means it's commands get promoted and we can reference these by command directly (dropping the plugin prefix).

```ts
{
  ...whatItWasBefore,

  'generate': Command,
  'help': Command,
  'lint': Command,
  'g': Command,
  'h': Command,
  'stories generate': Command,
  'stories g': Command,
  'stories reset': Command,
}
```

#### Other People's Plugins (O.P.P.)

Now we start calling `.plugin()` to add more commands:

```ts
{
  ...whatItWasBefore,

  'ignite backfill': Command,
  'mithril run': Command,
  'blueprint import': Command,
  'blueprint i': Command,
}
```

#### Default Command

Oh, and don't forget our default command.  This will run if we type nothing.  Let's make that run help.

```ts
{
  ...whatItWasBefore,

  '': Command,
}
```

We could even introduce the notion of a default command per-plugin quite easily.

```ts
{
  ...whatItWasBefore,

  'ignite': Command,
}
```

#### Our Final Map


```ts
{
  'am generate': Command,
  'am help': Command,
  'am lint': Command,
  'am g': Command,
  'am h': Command,
  'am stories generate': Command,
  'am stories g': Command,
  'am stories reset': Command,
  'generate': Command,
  'help': Command,
  'lint': Command,
  'g': Command,
  'h': Command,
  'stories generate': Command,
  'stories g': Command,
  'stories reset': Command,
  'ignite backfill': Command,
  'mithril run': Command,
  'blueprint import': Command,
  'blueprint i': Command,
  '': Command,
  'ignite': Command,
}
```

## Step 2: Find

We now need to `.find()` a command.  Let's say the user types this:

```sh
am generate mess everywhere --name Steve --dry-run
```

`context.parameters.array` gives us `generate mess everywhere`, so let's lookup it up in our map:

* `map['generate mess everywhere'] // undefined`
* Let's `shift` a word
* `map['generate mess'] // undefined`
* `shift` again
* `map['generate'] // Command` -- TIME TO SHINE BABY

What about:

```sh
am omg this might work
```

```ts
map['omg this might work'] //undefined
map['omg this might'] //undefined
map['omg this'] //undefined
map['omg'] //undefined
map[null || ''] // Command 
```

🤦‍♂️ on the fallback... but ...  🍒 🎰 WINNER 🎰 🍒 


## Feature Unlocking

It opens the door to `default` commands per-plugin.

Overriding something is as simple as clobbering an existing entry in the map.  This is mostly like a horrible idea.  However, it allows for some really interesting diverging options with in our so called `am` project.

jamonholmgren:
Thanks, @skellock, that's really well described.

I actually considered building this "every-possible-path" mapping structure when I was working this PR, but discarded it because it felt somewhat wasteful and I couldn't think of any real advantages over the current findCommand heuristics.

> It opens the door to default commands per-plugin.

This is possible already. Just name a command the same name as the plugin.

```
movie
  -- (gluegun cli here)
movie-imdb
  commands
    imdb.js
```

I tested it:

image

image

This should probably be documented (my next task: run through all the docs in preparation for 2.0.0 launch).

> clobbering an existing entry in the map

Is this something you need? I'd be interested to hear the use cases for this.

jamonholmgren:
Regarding the command knowing about its plugin smell ... I don't have a strong opinion on it. It was useful to have that information to not pass it around along with the command in a few key places that would be a pain to refactor. Since it's mostly internals, my preference would be to keep it simple and keep the reference on the command, but I could be convinced otherwise.

skellock:
Roger. Merge away!

skellock:
> Is this something you need? I'd be interested to hear the use cases for this.

Nah.

justsml:
@skellock brings up a great suggestion - it sounds like how some routing libs consolidate path+action mappings.

skellock:
Nice @justsml, I didn't really think of that.  That's a wonderful analogy.

It's easy to imagine how to build these new features with a route map.

### option-only commands

Such as:

* `xcodebuild -showsdks`
* `wget -V`
* `npm -l`

Allow people to flag options (somehow) as command-able and jam that in the map.

###  shell-based auto-complete

Search the map and `.filter()` results used by [tabtab](https://github.com/mklabs/node-tabtab)?

```js
function autocompleteFor(input) {
  const nextValues = Object
    .keys(routeMap)
    .filter(x => x.startsWith(input))
    .map(x => x.slice(input.length))
    .map(x => x.trim())
    .filter(x => x.length > 1)
    .map(x => x.split(' ').find(Boolean))
    .sort()

  return [...new Set(nextValues)]
}
```

###  VS Code plugin

If you show me the route map as json, I will show you how to make a VS Code extension so you can [launch tasks from a plugin](https://github.com/skellock/vscode-just/blob/master/src/vscode/execute-run-command.ts#L18-L37).

### Multi Ball Mode!!!

```sh
# ignite reset watchman
# ignite reset babel
# 
# maybe just run this instead:
ignite reset *
```

Use `*` to target all sub commands?  I bet that's asking for trouble.

### Socket Mode Engage

```sh
curl -s localhost:1336/generate/models?name=user&force=
```

Here, you map some kinda of janky url scheme to your routing system. (slashes are spaces, qsv's are option flags, lulz)

(This one I can show you a use case for tomorrow. I already wrote a 💩 version.)

### Conclusion

beat-a-dead-horse

Seriously though, I'm 👍  on this PR as is.  Maybe just revisit this the next time you need to address routing.

jamonholmgren:
Great discussion. I love the ideas. Let's keep this cooking in the backs of our brains for next time.

jamonholmgren:
I'm going to cut a new release and then work on documentation. Not tonight, unfortunately. I have an iPad acting up and have to drive to Portland to get it figured out. :P

justsml:
Thx Steve.

Jeez @jamonholmgren  - if your iPad is already sentient enough to make it to portland 😨 uh oh... 
Skynet is out & running on iPads?!?! We have just upgraded our problems...

jamonholmgren:
The new resolver in #352 is even better than is proposed here, IMO. I'm really happy with how it turned out.

Issue #331, opened 2018-01-18
Make nested commands work a little more intuitively


jamonholmgren:
@ryanlntn had some confusion around nested commands. He didn't know we needed the root command included in the folder in order for it to work.

Let's say you needed `foo stories generate` as a command.

```
commands/
  stories/
    generate.ts
```

This will fail, because it doesn't know what `foo stories` is.

However, if you add the root command:

```
commands/
  stories/
    stories.ts
    generate.ts
```

... it works, because now it'll crawl through `stories.ts` and keep going.

We can document that you need a root command, as above.

We could also assume that a folder is an empty no-op command and just add one if it doesn't exist. That would probably be a more intuitive solution, and not particularly hard to do.

Looking for feedback on this one.

skellock:
I like it. That makes a lot of sense.

ryanlntn:
Yeah I'm all for adding a no-op command if one doesn't exist.

skellock:
Creating a `folder` to get hierarchy is magical.

jamonholmgren:
Sounds good, let's do it.

jamonholmgren:
Closed by #352.

PR #352, opened 2018-02-14
🦅 Improvements to command path resolver

jamonholmgren:
The command resolver (`findCommand`) is simpler and better, and the result is the following two improvements to command resolution:

###  Plugins no longer automatically namespace, which allows for adding commands at the root level via plugin

Let's say you have a plugin that adds a few commands, called `movie-imdb`. It adds `movie imdb list` and `movie imdb search` commands.

In the past, you'd structure your plugin like this:

```
movie-imdb
  src
    commands
      list.js
      search.js
```

Now, this would result in the commands being `movie list` and `movie search`. To regain the `imdb` namespace, just move your commands into an intervening folder:

```
movie-imdb
  src
    commands
      imdb
        list.js
        search.js
```

This allows for a greater amount of freedom in how plugins affect your CLI, in that they can add root-level commands.

###  Nested commands can be resolved specifically by their folder structure, per #331.

In implementing the above change, I was able to also resolve issue #331, which now uses folder structure to resolve commands rather than requiring a command at every segment.

For example, in the past, to do `movie imdb tvshows search`, you'd have to do something like this:

```
movie
  src
    commands
      imdb
        imdb.js
        tvshows
          tvshows.js
          search.js
```

Now, you don't have to have the intervening commands:

```
movie
  src
    commands
      imdb
        tvshows
          search.js
```

`movie imdb tvshows search` will run the proper command now.

### DIFF
--- a/src/core-extensions/template-extension.test.ts
+++ b/src/core-extensions/template-extension.test.ts
@@ -9,13 +9,13 @@ const createRuntime = () => {
 }
 
 test('generates a simple file', async t => {
-  const context = await createRuntime().run('generate simple')
+  const context = await createRuntime().run('simple')
 
   t.is(context.result, 'simple file\n')
 })
 
 test('supports props', async t => {
-  const context = await createRuntime().run('generate props Greetings_and_salutations', {
+  const context = await createRuntime().run('props Greetings_and_salutations', {
     stars: 5,
   })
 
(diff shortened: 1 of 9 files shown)


GantMan:
Jamon this looks great, it's also perfect for a feature I want in Solidarity.

I love it!  Let me know when I can test it!

Issue #355, opened 2018-02-18
Improve startup performance


jamonholmgren:
Gluegun currently takes about 480ms to boot. That's compared to about 115ms for `node --version` on my machine.

I think we could improve startup performance by about 30-35% by disabling unneeded core extensions. The three that cause the most performance penalty are:

```
prompt +100ms, +4kb
print +45ms, +8kb
http +30ms, +0kb
system +10ms, +8kb
```

The rest are negligeable.

### API changes

The builder API would be something like this:

```js
  const cli = build()
    .brand('gluegun')
    .exclude(['semver', 'system', 'prompt', 'http'])
    .src(__dirname)
    .help()
    .version()
    .create()
```

I created a WIP branch and this cut Gluegun startup time from 480ms to 315ms, which is nontrivial. Removing the `.exclude()` call brought it right back up to 480ms, so it's working as intended. Disabling _all_ core extensions brought it down to 275ms, but that's mostly the `print` core extensions which is almost always needed.

### Downsides

1. We won't be able to use Rollup anymore. This is because Rollup needs to know whether to include the core extensions in the bundle or not. I tried a few things and concluded the best thing to do would be to go away from Rollup and ship an NPM package with all the files instead of just a rolled-up index.js
2. We won't be able to export the core extensions from gluegun proper anymore, for example, `import { prompt } from 'gluegun'` wouldn't work. To be honest, this has never felt very clean to me anyway, so it doesn't sadden me to see it go.

### WIP branch

Go here to see my WIP: https://github.com/infinitered/gluegun/compare/perf/startup-time

I'd like to hear thoughts about this, ideas, anything else.

skellock:
Ditching #2 is really bad news for me.  I use that everywhere.  

I disagree with the clean statement. The idiomatic way to import things in `js` and `ts` is via `import/require`.

Gluegun has two different ways to deliver functions to you, with & without context.  Perhaps this is the thing that feels a bit unclean to you (the fact there's two ways)?

Without direct exports, you're left shoehorning `Context` into every function call when all you need is a handful of `export { strings } from 'gluegun'`.  It gets ugly quick.

So perhaps some discussion around what `Context` should mean might be a good start?

It currently means: Tools + Context of the current user's request.

jamonholmgren:
[...]

PR #357, opened 2018-02-21
⚡️ Improves Startup Performance, Adds 'gluegun/toolbox' for on-demand loading

jamonholmgren:
Per #355, I decided to add a method to the builder to allow excluding certain core extensions if you don't need them and want to save the startup cost. `print`, `prompt`, `http`, and `system` are the slowest loading.

The API change simply adds this method, like this:

```js
  const cli = build()
    .brand('gluegun')
    .exclude(['semver', 'prompt', 'http', 'patching'])
    .src(__dirname)
    .help()
    .version()
    .create()
```

Additionally, these tools are not available via `require('gluegun')` anymore. However, they are exposed in `require('gluegun/toolbox')`.

```js
const { semver, prompt } = require('gluegun/toolbox')
```

Lastly, this pull request removes Rollup 😢because it no longer makes sense in the context of on-demand loaded resources.


jamonholmgren:
Merging without feedback :shame: ... trying to cut a new release soon, so I'm pushing forward.

skellock:
When I read notifications from my iPad, it marks it as read on github. I then have to remember to search for it from my laptop, or it drops off my radar.  :(

Never hesitate to @skellock me if you think I missed something.  Especially on our open source stuff.

If I don’t respond within 24 hours, send out a search party.

skellock:
I’m cool with this PR.

I find it so weird it takes that long to boot up a few of these libs.  Makes me want to dig in to find out why.

jamonholmgren:
> Makes me want to dig in to find out why.

Same, especially with `print`, because that one is pretty much _always_ necessary. I can understand `prompt` with all the plugins it has to load and whatnot.

skellock:
I wonder if it is the colors lib?  I know they monkeypatch the string prototype.  Perhaps that’s costly?  Perhaps chalk is better?

GantMan:
Jamon this new function is awesome.   When you generate a plugin should it have a commented out line with this in place, and everything listed?  And a comment?


So if I'm new, I generate a new project from your CLI generator, and now there's a friendly line that tells me (a total newb) how to make my startup faster?

jamonholmgren:
@skellock We can avoid monkeypatching string by using `'colors/safe'`.

https://github.com/Marak/colors.js#using-string-safe-api

I think we should make this change. It shouldn't affect any documented API. I'll make an issue.

@GantMan Yes, we should add that. I'll add an issue.

GantMan:
I'm eager to add these features to Solidarity.  When are you cutting a build?

GantMan:
bc Solidarity 2.0 is coming very soon

jamonholmgren:
@GantMan Getting close! Hopefully this week.

jamonholmgren:
@GantMan New Solidarity gluegun update here: https://github.com/infinitered/solidarity/pull/177

PR #358, opened 2018-02-24
👷 Refactors Builder to accept options in any order

jamonholmgren:
*This patch is backwards-compatible.*

This change refactors the `Builder` class to allow options in any order. This builds on #357 and solves one of the annoyances I had with the builder API (which I, unfortunately, introduced in #307), which is that it became order-dependent. This now has an internal `data` object that holds the info until everything is loaded.

The one exception to the order-dependent is still the `.brand()` method (this was a requirement prior to #307 too). So I added an optional `brand?: string` argument to the `build()` function to make it more idiomatic to specify the brand up front.

```js
// old way -- still works!
const cli = build()
  .brand('mycli')
  // ...
  .create()

// new way -- documented
const cli = build('mycli')
  // ...
  .create()
```

### DIFF
--- a/docs/runtime.md
+++ b/docs/runtime.md
@@ -2,13 +2,12 @@ With the gluegun API, you're able to load & execute commands.
 
 Check out the [sniff](./sniff.md) module for detecting if your environment is able to run.
 
-Here's what we're about to cover.
+Here's a kitchen sink version, which we're about to cover.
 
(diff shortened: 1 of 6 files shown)


jamonholmgren:
Sweet. It's feeling _goooooood_ now. Still have to work on some docs!

PR #373, opened 2018-03-12
🎛 Switches ToolboxParameters to GluegunParameters

jamonholmgren:
Minor naming tweak and fixing some TypeScript issues.

### DIFF
--- a/src/toolbox/print-tools.ts
+++ b/src/toolbox/print-tools.ts
@@ -1,9 +1,12 @@
 import * as CLITable from 'cli-table2'
-import * as colors from 'colors'
+import * as importedColors from 'colors'
 import { commandInfo } from './meta-tools'
 import { Toolbox } from '../domain/toolbox'
 import * as ora from 'ora'
 import { times, flip, prop } from 'ramda'
+
+// wtf typescript
+const colors: any = importedColors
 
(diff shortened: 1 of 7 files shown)


src/toolbox/print-tools.ts
```diff
@@ -1,10 +1,13 @@
 import * as CLITable from 'cli-table2'
-import * as colors from 'colors'
+import * as importedColors from 'colors'
 import { commandInfo } from './meta-tools'
 import { Toolbox } from '../domain/toolbox'
 import * as ora from 'ora'
 import { times, flip, prop } from 'ramda'
 
+// wtf typescript
+const colors: any = importedColors
```
skellock: lol!

jamonholmgren: That was like 20 minutes of frustration wrapped into one comment

skellock: What was the problem?  It looks like you may have been trying to redeclare a variable name.

jamonholmgren: It wasn't the redeclaring (that was the solution I eventually landed on, so I could make it be typed `any`). It was because the TS types weren't following `colors` around properly. I'll take another look at it eventually.

PR #376, opened 2018-03-15
🥌 Removes gluegun/toolbox, refactors core extension loading

jamonholmgren:
I removed the `gluegun/toolbox` and refactored the way core extensions are loaded so they can be loaded directly.

Before, we'd expose extensions at `require('gluegun')` (and subsequently at `require('gluegun/toolbox')` by doing something like this:

```js
import attachStringExtension from './core-extensions/string-extensions'

const fakeContext = {}
attachStringExtension(fakeContext)
const strings = fakeContext.strings
export { strings }
```

This is kind of funky and precludes tree shaking and other optimizations. So I refactored the way core extensions are loaded, which is more like this:

```js
export { strings } from './toolbox/string-tools'
```

In the core extension itself, we attach it like this:

```js
import { strings } from '../toolbox/string-tools'
function attach(context) {
  context.strings = strings
}
```

So in both cases, we're pulling in the tool with import/export. With the extension, we attach it to the context, and with the on-demand toolbox, we just export it immediately.

Note: I wasn't able to export `meta` or `template` due to how they rely on having a RunContext. Open to ideas.


skellock:
> Note: I wasn't able to export meta or template due to how they rely on having a RunContext. Open to ideas.

I'd say don't sweat it.  My primary use case was getting access to the string tools and such.  

One that's unclear to me here, can I do this:

```ts
import { pascalCase } from 'gluegun/strings'
import { read } from 'gluegun/filesystem'
import { patch } from 'gluegun/patching'
```

jamonholmgren:
Not *yet*. I will add that in the next PR. Should be super easy, but I want to write more robust tests for it.

PR #387, opened 2018-09-14
fix(system-toolbox): Fix spawn and run reject

cmckni3:


src/toolbox/system-tools.ts
```diff
@@ -75,7 +76,7 @@ async function spawn(commandLine: string, options: Options = {}): Promise<any> {
     })
     spawned.on('error', err => {
```
cmckni3: I believe there may be a race condition in the lines above as well

PR #388, opened 2018-09-15
ci(azure-pipelines): Add Azure pipelines config

### LINKED ISSUE #386: CI should be setup for Windows to avoid breaking support in the future
As mentionned by @cmckni3, CI should be setup for Windows as well.
This would avoid breaking Windows support in the future.

- I can help with the Azure Pipeline setup
- Someone from infinitered would need to
    - Setup the Microsoft account
    - Setup the checks for PRs in github

cmckni3:
Closes #386


cmckni3:
WIP fixing cli.integration.test.ts path issues

azure-pipelines.yml
```diff
@@ -0,0 +1,20 @@
+pool:
+  vmImage: 'vs2017-win2017'
```
chrisrpatterson: this will need to be vs2017-win2016

azure-pipelines.yml
```diff
@@ -0,0 +1,20 @@
...
+
+steps:
+- task: NodeTool@0
+  inputs:
+    versionSpec: '10.x'
+  displayName: 'Install Node.js'
+
+- script: |
```
chrisrpatterson: On Windows -script is going to run with CMD as a batch file and not with bash.  You can try -bash which should run with the bash included with msys embedded in Git

azure-pipelines.yml
cmckni3: Do you have a reason why? Looks like vs2017-win2016 has .NET Core (which this project does not use) and vs2017-win2017 does not?

https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=vsts&tabs=yaml

azure-pipelines.yml
cmckni3: Thanks! Seems to function much smoother with `- bash`

cmckni3:
This works in my test Azure pipeline. The integration tests are very finicky on Windows though.

kevinvangelder:
Looks like Azure pipeline was queued but hasn't actually run (under the Checks tab). Any suggestions?

cmckni3:
My guess is that this may have to be merged first and then select `azure-pipelines.yml` in the pipeline UI.

cmckni3:
image

chrisrpatterson:
If you set the pipeline up initially through the app flow PR builds should automatically be enabled.  If you setup through the OAuth flow then you might need to go and enable the PR trigger on the triggers tab.

Please let me know if you are not able to get it working.

cmckni3:
Thanks for pointing that out @chrisrpatterson. My test project was created using the OAuth flow and the push triggers were added automatically. I believe there is a toggle as you mentioned to enable PR builds.

kavengagne:
Please note that the hosted virtual machines used for Azure Pipelines are adding the git bin folder to the PATH, which enables tools like CP and RM, which are not normally present on Windows developers machines.

This means that even if the CI build passes on Azure Pipelines, it is not guaranteed to work on a Windows developer machine.

cmckni3:
Yeah those will have to be migrated to JS as well.

jamonholmgren:
I'm going to wait on this one until post-2.0 release.

RichiCoder1:
My PR https://github.com/infinitered/gluegun/pull/419 might help with this. Removed bash-isms in tests and npm scripts.

jamonholmgren:
@cmckni3 Want to bring this up to date?

cmckni3:
Merged master and updated the azure ci command to `yarn ci:test`

chrisrpatterson:
Is there anything we can do to help out here?

cmckni3:
I don’t have visibility (read: permissions)to connect the repo to Azure DevOps.

@chrisrpatterson can you review the pipelines file? Also, any suggestions for testing with a base Windows image without `git` and `cp` added to PATH?

chrisrpatterson:
@cmckni3 Generally it looks fine to me.  I might install yarn with ```choco install yarn```  instead of using the bash script.  

For the other issue, I am not sure what you could do outside of using something like grunt or make.js to run your build actions and use the various node modules in those to abstract things like copy commands.  Yarn and NPM scripts don't seem to have a good way to handle needing slightly different tools on different platforms.

RichiCoder1:
> @cmckni3 Generally it looks fine to me. I might install yarn with choco install yarn instead of using the bash script.

Agreed.

> For the other issue, I am not sure what you could do outside of using something like grunt or make.js to run your build actions and use the various node modules in those to abstract things like copy commands. Yarn and NPM scripts don't seem to have a good way to handle needing slightly different tools on different platforms.

I'd npm run scripts are perfectly fine at this point. Most of the commands that need cross plat equivs have packages now that are super easy to use in run scripts. I already did the hard work of de-platformify the run scripts :)

cmckni3:
Yeah I originally used script instead of bash in the azure devops config.

RichiCoder1:
`&&` chaining works in batch and bash, but not powershell. However I have been running my default shell as powershell and I still haven't had issues so *shrug*. And I'm pretty sure the issues I ran into were just because I didn't pull your `cli-intergration` test fixes.

src/cli/cli.integration.ts
```diff
@@ -2,14 +2,15 @@ import * as expect from 'expect'
...
+import * as path from 'path'
 import { run as cli } from './cli'
 
 sinon.stub(console, 'log')
 
 const pwd = process.cwd()
 
 // set jest timeout to very long, because these take a while
-beforeAll(() => jest.setTimeout(90 * 1000))
+beforeAll(() => jest.setTimeout(180 * 1000))
```
cmckni3: I'm trying this back out as `90`. CI seems a bit slow with this long of a delay.

cmckni3: Appears to be necessary in Windows

cmckni3:
Interesting. Didn't know it worked in batch.

cmckni3:
Ping @jamonholmgren 

Let me know if you help setting it up in the Azure DevOps Portal.

Steps I used:

1. Create Azure DevOps Project
2. Builds > New Build Pipeline
3. Select GitHub for Repository > Authorize GitHub
4. Select infinitered/gluegun from `...` menu or input manually
5. Select branch
6. Hover over "Configuration as Code > YAML" > Click Apply
7. Give the pipeline a name or leave default
8. Select "Hosted" as the Agent Pool
9. Select azure-pipelines.yml for Yaml file path
10. Click "Save & queue"

chrisrpatterson:
I am trying to add reporting for test results only or CI but Jest does not allow you to provide reporter options on the command line.  The result is junit.xml getting overwritten by the integration tests after the unit tests run. 

Without either splitting integration tests out of the ci:test target or having different configuration for CI and local run I am out of ideas.

chrisrpatterson:
@RichiCoder1 awesome that did the trick https://dev.azure.com/chrispatmsft/GitHubProjects/_build/results?buildId=2419&view=ms.vss-test-web.build-test-results-tab.  Unfortuately the CI is failing for me at the moment so I will need to figure that out

jamonholmgren:
Apologies for the radio silence. This is now on my radar and I'll be looking into it soon!

cmckni3:
Ah never added the test result publishing. Is that something you're interested in @jamonholmgren? Does that happen in circleci at the moment?

cmckni3:
Nice to see the status checks for Azure Devops!

jamonholmgren:
I'm not even sure what "test result publishing" means, @cmckni3 !

jamonholmgren:
I'm not sure why Azure ran twice. Perhaps because I started with a jamonholmgren/gluegun org and then deleted that and moved it to infinitered/gluegun in Azure DevOps.

PR #395, opened 2018-12-03
Fix command name printing

yordis:
I have the following command

```
module.exports = {
  name: 'app.build',
  dashed: false,
  description: 'Build an application',
  run,
};
```

Under the file `build.js`

When I run `yarn cli --help` the output shows me `build` as an available command but this is not true, the command name is actually `app.build`

### DIFF
--- a/src/toolbox/meta-tools.ts
+++ b/src/toolbox/meta-tools.ts
@@ -59,10 +59,7 @@ export function commandInfo(toolbox: GluegunToolbox, commandRoot?: string[]): st
     }),
     map(command => {
       const alias = command.hasAlias() ? `(${command.aliases.join(', ')})` : ''
-      return [
-        `${command.commandPath.join(' ')} ${alias}`,
-        replace('$BRAND', toolbox.runtime.brand, command.description || '-'),
-      ]
+      return [`${command.name} ${alias}`, replace('$BRAND', toolbox.runtime.brand, command.description || '-')]
     }),
   )(toolbox.runtime.commands)
 }


jamonholmgren:
@yordis I wonder if this might break if there are nested commands. Would you mind writing a few tests for it?

jamonholmgren:
Like this:

```
moviecli
  commands
    foo
      bar
        bar.js
```

```
$ moviecli foo bar
$ moviecli --help
```

Will this still show `foo bar` or will it only show `bar` in help?

yordis:
@jamonholmgren without actually testing it, it shouldn't matter unless I am misunderstanding something.

That value comes from the configuration on your commands https://github.com/infinitered/gluegun/blob/5cd228e7d98b02354b231e16a78a3260625e999e/src/domain/command.ts#L6

Why would the folder structure imply the command name instead of the actual `name` key?

jamonholmgren:
The problem is that the `commandPath` contains the whole path to the name, where your solution only includes the name. You can see it when you output help:

image

Notice there it shows `foo bar` ? If I use your version, I get this:

image

Only `bar` now. This won't work, because the real way you invoke the command is `foo bar`, not `bar`.

yordis:
@jamonholmgren wait so are you telling me that having nested commands will affect my command name?

For example today I have `app.build` because and only because I want to be scoped by `app.`.

How do you structure your folders and so on? I missed the documentation about this. Maybe the true output is `the command  + command name`

jamonholmgren:
@yordis I should make the nested commands documentation better, but yeah: https://github.com/infinitered/gluegun/blob/master/docs/guide-architecture.md#commands-file-structure

So, if you have a `movies` CLI:

```
commands
  foo
    bar
      baz.js
```

will let you do `movies foo bar baz` and run the `baz.js` command.

jamonholmgren:
@yordis It's not strictly necessary, we could go by the name of the file instead. In fact, if you leave off the name, it will "guess" the name by the name of the file. So the `name` property is optional.

Actually, you've found a bug, now that I've tested it. If I change the `name` it still shows the wrong name in `help`. But it does work if you type it in. Want to take a stab at fixing that bug, @yordis ?

yordis:
@jamonholmgren yeah, give me some insides of what you did, more than happy to fix it, that way I learn more about gluegun.

jamonholmgren:
@yordis Just spin up a new Gluegun app:

```
gluegun new myapp
```

Add a new command in `src/commands/foo/bar/baz.js:

```
module.exports = {
  name: "yordis",
  run: async toolbox => toolbox.print.info("hi!")
}
```

Then run `yarn link` and `myapp --help`. The help will show the wrong name, not `foo bar yordis` but rather `foo bar baz`.

yordis:
@jamonholmgren tested and fixed

PR #396, opened 2018-12-03
Remove unnecessary information from argvs

yordis:
From the following command

```
straw-hat test --watch
```

Previously `argv` will give me the exact same information as `raw` would give me.

Based on my needs, I will have to do `.slice(3)` in `raw` or `argv` so I propose to this by default on `argv`

Here is the argument for it.

- The first  position of the list will be aways `node`
- The second of the list will be always the name of the CLI `straw-hat`
- The third position of the list will be always the command name `test`

And none of that information is useful or relevant inside the commands since you wouldn't need to get that info from that list anyway (this is an static information that you know upfront already)

cc: @jamonholmgren I hope you remember about this


jamonholmgren:
To be safe, it would probably be better to just add another property to the `parameters` object. Perhaps `arguments`?

yordis:
@jamonholmgren yep

yordis:
@jamonholmgren fixed

src/runtime/run.ts
```diff
@@ -41,6 +41,7 @@ export async function run(this: Runtime, rawCommand?: string | string[], extraOp
     options: toolbox.parameters.options,
     raw: rawCommand,
     argv: process.argv,
+    arguments: process.argv.slice(3),
```
kevinvangelder: If we were to include this I would think we would want this number to be configurable by the GlueGun consuming author. Additionally, can you reasonably expect the `--watch` to be the third option and not the Nth option?

yordis: @kevinvangelder I am just remove the information that we already know about the command

- The first position of the list will be aways node
- The second of the list will be always the name of the CLI straw-hat
- The third position of the list will be always the command name test

I am not following why would you make that configurable

src/runtime/runtime-parameters.test.ts
```diff
@@ -18,6 +18,19 @@ test('can pass arguments', async () => {
   expect(command.commandPath).toEqual(['hello'])
 })
 
+test('can pass the normalized arguments', async () => {
+  const r = new Runtime()
+  const { parameters } = await r.run('test --watch --coverage')
+  const { arguments: argvs } = parameters
```
jamonholmgren: I just realized `arguments` is a reserved keyword. Any other ideas?

src/runtime/runtime-parameters.test.ts
```diff
@@ -18,6 +18,19 @@ test('can pass arguments', async () => {
...
+
+  //  { array: [ 'test' ], options: { watch: true, coverage: true } }
```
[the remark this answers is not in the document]
jamonholmgren: That looks right to me, other than missing some things. What do you find as strange here?

src/runtime/runtime-parameters.test.ts
yordis: @jamonholmgren OMEGALUL I that makes sense hahahahaha

I would use `argv` 🤷‍♂️  

I know that it may introduce breaking changes but with good documentation about it should be enough.

- If you were doing the slice, then they will no need to do it anymore
- If they wanted the raw argvs then they should use `raw` instead.

But, that is what I would do.

Give me directions on what you want to do.

src/runtime/runtime-parameters.test.ts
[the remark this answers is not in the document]
jamonholmgren: @yordis Ah yes, you were missing this:

```js
  r.addCoreExtensions()
  r.addPlugin(`${__dirname}/../fixtures/good-plugins/args`)
```

I'm adding it and fixing some things, will have a commit for here soon.

src/runtime/run.ts
jamonholmgren: So I think @kevinvangelder is right. The third position could be the command, but it could also be a nested command. So we can't be sure.

jamonholmgren:
@yordis I'm going to sit on this one and not include it in the Gluegun 2.0 release. I think we need to do more thinking about the ramifications of it.

yordis:
@jamonholmgren should I close this?

I believe that you have enough information on what to do and the changes required are really simple, maybe we should move the conversation to an Issue and figure out the direction.

jamonholmgren:
@yordis Yeah, let's close it for now. Thanks!

PR #397, opened 2018-12-03
Fix system.which command

ruddell:
If the command doesn't exist, `system.which` throws an error.  Based on the comments and return values, it shouldn't throw an error.  An example of where this is used and fails (if `git` not found): [Bowser Boilerplate](https://github.com/infinitered/ignite-ir-boilerplate-bowser/blob/master/boilerplate.js#L232)

[npm which docs](https://www.npmjs.com/package/which)

I added a test checking for `git` I'm guessing is present on any machine running the tests.  If anyone has a better idea, let me know

### DIFF
--- a/src/toolbox/system-tools.test.ts
+++ b/src/toolbox/system-tools.test.ts
@@ -2,7 +2,7 @@ import * as expect from 'expect'
 import { system } from './system-tools'
 
 test('which - existing package', () => {
-  const result = system.which('git')
+  const result = system.which('node')
   expect(result).not.toBe(null)
 })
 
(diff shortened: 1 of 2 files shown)


src/toolbox/system-tools.test.ts
```diff
@@ -0,0 +1,12 @@
+import * as expect from 'expect'
+import { system } from './system-tools'
+
+test('which - existing package', () => {
+  const result = system.which('git')
```
kevinvangelder: Maybe testing against `node` would be better, since Jest tests require it?

ruddell: Great idea

Issue #404, opened 2018-12-17
Update the dependencies in generated default package


markusguenther:
The dependencies in the package.json.ejs file are really outdated and should be adjusted to save the step of upgrading everything first.

PR #405, opened 2018-12-17
Update package json template

markusguenther:
The dependencies in the generated package are bit outdated and could be upgraded easily.

Resolves: #404


jamonholmgren:
Thanks so much @markusguenther! This will be shipped in 2.0.0.

Issue #406, opened 2018-12-18
Use jest instead of Ava in CLI template


markusguenther:
I know that ignite used Ava in the past and obviously gluegun also used it in the past.
Probably this should also be adjusted to jest. Gluegun it self uses jest at the moment, so why the generated package should base on ava.

PR #407, opened 2018-12-18
feat: Replace ava in favor of jest

markusguenther:
The generated packages still use ava and not jest

Resolves: #406


jamonholmgren:
Thanks a ton @markusguenther !!

PR #412, opened 2018-12-20
Upgrade all dependencies

jamonholmgren:
Was not able to automatically update `enquirer` and `fs-jetpack`. I'll have to manually do that at a later date.


src/loaders/command-loader.ts
```diff
@@ -31,8 +31,8 @@ export function loadCommandFromFile(file: string, options: Options = {}): Comman
   command.name = head(split('.', (filesystem.inspect(file) as any).name))
 
   // strip the extension from the end of the commandPath
-  command.commandPath = (options.commandPath || last(file.split('commands' + path.sep)).split(path.sep)).map(
-    f => ([`${command.name}.js`, `${command.name}.ts`].includes(f) ? command.name : f),
+  command.commandPath = (options.commandPath || last(file.split('commands' + path.sep)).split(path.sep)).map(f =>
+    [`${command.name}.js`, `${command.name}.ts`].includes(f) ? command.name : f,
```
jamonholmgren: Formatting only from here on down.

cmckni3:
My only comment is that I like using `^<major>.0.0` in package.json for 1.0 and above. < 1.0 would be `~0.<minor>.0`

jamonholmgren:
Unfortunately, Enquirer 2.0 is no longer compatible with many of the plugins we were using. It will be a pain in the butt (I just tried) to upgrade. So I'm going to lock at 1.0.3 and we'll stay there for a while.

PR #413, opened 2018-12-20
Update readme.md

noneissome:
Fixed readme


cmckni3:
LGTM but I would say 8+. Specifically 8 or 10 to encourage using LTS versions of node.

cmckni3:
Maybe a note somewhere also to say it will work with >= 7.6 but LTS versions of node are recommended.

Thoughts @jamonholmgren?

jamonholmgren:
@cmckni3 I agree with you, although this PR is good to merge now so I will. Maybe @noneissome can do another PR with that.

noneissome:
> @cmckni3 I agree with you, although this PR is good to merge now so I will. Maybe @noneissome can do another PR with that.

Shall do.

PR #414, opened 2018-12-20
Typescript improvement to CLI commands

jamonholmgren:
Generated TypeScript CLIs will now have the most basic `GluegunToolbox` types added, since this can be unintuitive.


jamonholmgren:
@cmckni3 Because of the way we're loading up the `toolbox` at runtime, TypeScript fails to infer it. I figured this out when I was doing some test CLIs. So we do need this here.

cmckni3:
@jamonholmgren good point

PR #415, opened 2018-12-20
Updates fs-jetpack to latest

jamonholmgren:
So there's one sticking point with this pull request. fs-jetpack [removed the symlinks](https://github.com/szwacz/fs-jetpack/blob/master/CHANGELOG.md#200-2018-07-10) option, which is unfortunate given that it breaks backwards compatibility for us, too.

Maybe this is dumb, but I really don't want to release a 3.0 so soon after 2.0, especially for such a tiny change.

So, my thought is to keep this PR open and updated, and when we're ready to deploy a 3.0 sometime down the road, we merge this in.

The major benefit from the latest version is we get built-in TypeScript types from fs-jetpack itself rather than rolling our own.

Thoughts? @cmckni3 @GantMan


GantMan:
One interesting aspect is this:  Are breaking changes on tools considered major release worthy?  If so, that's a lot of major releasing.

Maybe major release for Gluegun API changes, and Minor releases (in this case 2.1.0) for toolbox updates.

jamonholmgren:
@GantMan Yeah, that's a really interesting thought. Since we're exposing other libraries as our own (and sort of hiding them kinda but not really), their APIs are our APIs. So for strict semver we'd need to bump major versions every time. But honestly, that seems like too much for me. I'd be okay with making a judgment call on this.

cmckni3:
I agree with a major release of gluegun for this. It's annoying to have another major release for this. The missing option won't catch gluegun users by surprise with a major update. Of course, as long as it's documented somewhere (such as `CHANGELOG.md`).

PR #419, opened 2018-12-21
Adds types to extensible toolbox and fixes build and test on Windows

### LINKED ISSUE #418: Typescript Inconsistencies
Was referred to this (excellent) library by a coworker and ran into a few small issues while building out a typescript.

Biggest one was passing `GluegunCommand` a generic paramter with a customized `Toolbox` (for inbox helpers, etc) results in losing typings on all the `Gluegun` in-built helpers as the exported `Toolbox` type which you'd inherit from has a quirk that kills all types for the implements base. I'm pretty sure this is easy to fix though, and will submit a PR. `Toolbox` is also missing some typings.

It may also be work generating a `{{appName}}-types.d.ts` with an `export interface {{appName}}Toolbox extends Toolbox` and `export declare {{appName}}Command = GluegunCommand<{{appName}Toolbox>` for the typescript template as a starting point for peeps who might want to type helpers.

RichiCoder1:
This, uh, blew up a little.

Originally it was just about correcting the types for `Toolbox` (which is does) but I ran into a bunch of issues building and running tests on Windows and so I ended up fixing most of those issues

Partially resolves #418


jamonholmgren:
Hey @RichiCoder1 , thanks so much for your hard work on this!

Looks like CI tests are failing with

```

 FAIL  src/core-extensions/strings-extension.test.ts
  ● Test suite failed to run

    ENOMEM: not enough memory, read

      at ScriptTransformer._transformAndBuildScript (node_modules/jest-runtime/build/script_transformer.js:371:53)
      at Object.<anonymous> (node_modules/ramda/src/index.js:19:39)
```

Any ideas? I assume they're passing on your end?

jamonholmgren:
image

The tests (`yarn ci:test`) completed successfully on my machine. Perhaps there's a way to reduce the memory footprint somehow so the CI doesn't blow up?

RichiCoder1:
Also passing here. Not sure what would be causing the memory issues? Can investigate

buildHelpers/cleanDir.js
```diff
@@ -0,0 +1,31 @@
+const { readdirSync, statSync, rmdir, unlink, existsSync } = require('fs')
```
jamonholmgren: Couple things here:

1. Can you run this through prettier?
2. I'm kind of loath to add another folder just for this file. Is there somewhere else we could put this?

RichiCoder1: 1. Sure! Sorry, forgot it doesn't get automatic coverage.
2. I had the same thought, but wasn't sure where to put it. I'm def open to ideas. Sadly couldn't reproduce it consiely in nom scripts.

jamonholmgren: Let's call the folder `ci` and then I'll add some more things in there that I think will clean up the folder structure too.

jamonholmgren: Actually, hold on ... we already have a `.circleci` folder, and that might be confusing. Thinking...

jamonholmgren: I suppose there's no reason we can't put this script in `.circleci`. That's where it is being run, anyway. Let's just do that.

RichiCoder1: Ha, sounds good.

RichiCoder1: Moved and prettier'd!

RichiCoder1:
Fixed the build issue, but I did it by throttling works on CI. Doesn't seem the slow down the actually test run much. Thoughts?

jamonholmgren:
Works for me!

Also, I added a couple more build-time scripts via #426. Thanks for sparking the idea.

RichiCoder1:
Happy to help! Let me know if you have any other feedback, otherwise :shipit:

jamonholmgren:
This looks good. Shouldn't be any backwards incompatibilities that I'm aware of. I'll release as 2.0.3.

PR #420, opened 2018-12-22
Add documentation for `ts-node` and declaration files

yordis:
I made a mistake by using `master` for developing https://github.com/infinitered/gluegun/pull/395 🤦‍♂️ 

So ignore the extra stuff the real fix is for the template.

Just some documentation that will prevent people from researching for quite some time why the application was failing (ask me how I know 😢)


jamonholmgren:
Really appreciate you contributing back, @yordis!

PR #421, opened 2018-12-22
Fixes issues with movie tutorial

jamonholmgren:
I went through the tutorial and there were problems, so I fixed them (and a few other docs issues).


mikeerickson:
@jamonholmgren Thank you very much for doing this!

PR #426, opened 2018-12-23
Removes direct access files from repo, adds to build pipeline instead

jamonholmgren:
This PR will remove direct access files from the repo but generate them with the `yarn build` command. This cleans up the repo but keeps backwards compatibility.

Direct access are files that enable this:

```js
const { print } = require('gluegun/print')
```

cc @yordis who has asked for this in the past and @Richicoder1 who sparked the idea.

(Note I am keeping the build scripts in `./.circleci`, which I think is a reasonable place for things that are really only run by CircleCI)


RichiCoder1:
Nice, looks awesome! And cleaner initial pull :)

PR #428, opened 2018-12-23
Yordis/fix 378

### LINKED ISSUE #378: parameters.options type is Options | undefined
so when I write code like
```js
const { options } = parameters
``` 

I get issues because options COULD be undefined.  Instead, make options an object no matter.

yordis:
Fixes #378


jamonholmgren:
Thank you @yordis !!!

PR #435, opened 2018-12-28
#430 remove lock files from doc gitignore

### LINKED ISSUE #430: gitignore file lists lock files
The [.gitignore](https://github.com/infinitered/gluegun/blob/master/src/cli/templates/cli/.gitignore.ejs#L6) file lists down `package.lock.json` and `yarn.lock.json`.

As shown in the npm docs, this is a file that should be committed. 

Is there a reason for this?

BoyWithSilverWings:
Fixes #430 

This removes `lock` files from cli templates that are generated from user generated template. 
Since this is for the code that is generated for the user, both files should not be in `.gitignore` giving the user option to use either `yarn` or `npm`

#434 fixes this in the global package.

### DIFF
--- a/src/cli/templates/cli/.gitignore.ejs
+++ b/src/cli/templates/cli/.gitignore.ejs
@@ -3,5 +3,3 @@ node_modules
 npm-debug.log
 coverage
 .nyc_output
-yarn.lock
-package.lock


jamonholmgren:
Will publish with the next point release. Thank you!

PR #437, opened 2018-12-28
Improves performance - lazy requires

### LINKED ISSUE #391: Improve plugin loading performance
## Problem: slow startup when using lots of plugins

Per a discussion with some of the AWS Amplify folks, startup performance when using lots of plugins can be quite slow. They report 3-9 second startup times. This obviously won't work very well.

## Why it's happening

We are [eager-loading](https://github.com/infinitered/gluegun/blob/daf37d09f7524f6d99ae8ec2e38050ebd531b798/src/loaders/plugin-loader.ts#L49-L68) every extension and every command from every possible plugin. This results in file access and JavaScript parsing load times for many commands. Since Gluegun generally only needs to run one command at a time, all of that extra parsing is wasted.

#### 1. Add the ability for each plugin to decide whether to be loaded or not

When Gluegun finds a plugin, it could check for a `plugin.js` file at the root of the plugin directory. This is similar to [what Ignite does](https://github.com/infinitered/ignite/blob/9ebb23514284ee44366d885cceeb15487d9b57fc/src/commands/add.js#L117-L118) with its plugin architecture.

We can `require` that file and look for an exported `shouldLoad` function. Pass in the context (which will include the current command's parameters) and let the plugin itself decide whether to load or not by returning `true` or `false`.

Example:

```js
// mycli-foo/plugin.js
function shouldLoad (context) {
  // only load if the first parameter is `foo`, as in `mycli foo bar --baz=true`
  return (context.parameters.first === 'foo')
}

module.exports = { shouldLoad }
```

If the plugin decides it's not relevant to the current running request, it can decline to be loaded, thus speeding up the startup process.

#### 2. Intelligent command loader by scope

Our plugin loader could look at the passed-in arguments and try to determine which plugin needs to be loaded automatically based on the first argument, which would "scope" it. For example:

```
$ movies imdb search "Tom Hanks"
```

In this case, the loader could intelligently ignore all plugins except the `movies-imdb` plugin, because it starts with `imdb`.

This would be less flexible than the first option but also be easier. We might include an opt-in flag for backwards compatibility:

```js
const movieCLI = build('movie')
  .src(`${__dirname}/core-plugins`)
  .plugins('node_modules', { matching: 'movie-*', scoped: true })
```

However, this wouldn't allow plugins to add root-level commands, like if `movies-imdb` wanted to add a root-level `movies search` command. It also wouldn't allow for aliases like `movies i search`

## Discussion

I'd like to hear different use-cases and any other ideas people have.

jamonholmgren:
This PR improves performance and reduces startup time by about 1/3 by lazy loading a bunch of third-party packages.

Before PR: 482ms startup
After PR: 311ms startup

Fixes #391.


sniff.js
```diff
@@ -1,13 +1,13 @@
 // check the node version
-const semver = require('semver')
```
GantMan: Was semver a slowdown?   Good call on this then!

src/toolbox/print-types.ts
```diff
@@ -1,9 +1,20 @@
...
+
+export type GluegunPrintColors = typeof importedColors & {
+  highlight: (t: string) => string
+  info: (t: string) => string
+  warning: (t: string) => string
+  success: (t: string) => string
+  error: (t: string) => string
+  line: (t: string) => string
+  muted: (t: string) => string
+}
```
GantMan: 👍

src/toolbox/prompt-tools.ts
```diff
@@ -1,45 +1,48 @@
...
-enquirer.register('rawlist', promptRawlist)
-enquirer.register('confirm', promptConfirm)
-enquirer.register('expand', promptExpand)
-enquirer.register('checkbox', promptCheckbox)
-enquirer.register('radio', promptRadio)
-enquirer.register('password', promptPassword)
-enquirer.register('question', promptQuestion)
-enquirer.register('autocomplete', promptAutocompletion)
+let enquirer = null
+function getEnquirer() {
```
GantMan: Good call on switching to a singleton here.

sniff.js
jamonholmgren: @GantMan It was more significant than I expected! Since this file isn't super critical these days, I went to a simpler version check.

PR #444, opened 2019-01-09
Strengthens GluegunToolbox type guarantee

jamonholmgren:
Before this commit, `GluegunToolbox` had as optional various core extensions. This makes the core extensions non-optional (once the Toolbox is built) and introduces two new internal types, the interface `EmptyToolbox` and the class `GluegunEmptyToolbox`. These aren't intended to be used by CLI authors and not exported publicly.

This should be backwards-compatible, but my intention is to bump to 2.1.0 (as documentation pre-emptively says).

*Update:* There is one thing that will be not backwards-compatible, but is undocumented behavior -- if you didn't specify a default command, before, it would just quit without output. Now it throws an error. I had to update the `gluegun` CLI to include a default command.

### DIFF
--- a/src/cli/templates/cli/.gitignore.ejs
+++ b/src/cli/templates/cli/.gitignore.ejs
@@ -3,3 +3,5 @@ node_modules
 npm-debug.log
 coverage
 .nyc_output
+dist/**/*
+build/**/*
--- a/src/cli/templates/cli/.gitignore.ejs
+++ b/src/cli/templates/cli/.gitignore.ejs
@@ -3,5 +3,5 @@ node_modules
 npm-debug.log
 coverage
 .nyc_output
-dist/**/*
-build/**/*
+dist
+build
--- a/src/cli/templates/cli/package.json.ejs
+++ b/src/cli/templates/cli/package.json.ejs
@@ -10,6 +10,8 @@
     <% if (props.typescript) { %>
       "format": "prettier --write **/*.{js,ts,tsx,json}",
       "lint": "tslint -p .",
+      "compile": "tsc -p .",
+      "build": "yarn format && yarn lint && yarn compile",
     <% } else { %>
       "format": "prettier --write **/*.{js,json} && standard --fix",
       "lint": "standard",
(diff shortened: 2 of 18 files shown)


src/cli/templates/cli/.gitignore.ejs
```diff
@@ -2,4 +2,6 @@
 node_modules
 npm-debug.log
 coverage
-.nyc_output
\ No newline at end of file
+.nyc_output
+dist/**/*
+build/**/*
```
GantMan: I'm pretty sure you can just set these as `dist` and `build` and that means all files.  Use globs only when there is a subset of the folder to ignore.  And even then, if it's a single exception they can use `!`.

src/cli/templates/cli/package.json.ejs
```diff
@@ -10,6 +10,8 @@
     <% if (props.typescript) { %>
       "format": "prettier --write **/*.{js,ts,tsx,json}",
       "lint": "tslint -p .",
+      "compile": "tsc -p .",
+      "build": "yarn format && yarn lint && yarn compile",
```
GantMan: 👍

src/cli/templates/cli/readme.md.ejs
```diff
@@ -2,6 +2,25 @@
...
+
+```shell
+$ npm login
+$ npm whoami
+$ npm lint
+$ npm test
+(if typescript, run `npm run build` here)
+$ npm pack
+$ npm publish <name of tgz file here>
+$ rm <name of tgz file here>
```
GantMan: research if npm pack is necessary, I might just go with the simple one here.

src/runtime/runtime-find-command.ts
```diff
@@ -23,9 +23,16 @@ export function findCommand(runtime: Runtime, parameters: GluegunParameters) {
   let tempPathRest = commandPath
   let commandPathRest = tempPathRest
 
+  // a fallback command
+  const commandNotFound = new Command({
+    run: (toolbox: GluegunToolbox) => {
+      throw new Error(`Couldn't find that command, and no default command set.`)
```
GantMan: 🎆

PR #449, opened 2019-01-15
Support async extension setup

RichiCoder1:
Allows an extensions to return a promise on setup to allow for async operations like async FS IO, or remote fetching.


jamonholmgren:
@RichiCoder1 Thank you! Can you add some documentation on how to use this and a test? I would suggest documenting in this file: https://github.com/infinitered/gluegun/blob/0bad0f6fc3f3523f0ed5ca06c570ec2cf36da6c3/docs/plugins.md#extensions

Tests in here: https://github.com/infinitered/gluegun/blob/0bad0f6fc3f3523f0ed5ca06c570ec2cf36da6c3/src/loaders/extension-loader.test.ts

RichiCoder1:
Added tests and docs. Also added a timeout, but I'm half tempted to tear it out. Thoughts?

RichiCoder1:
Pulled it out! Let me know whatcha think

jamonholmgren:
I think we just need some documentation and then we're good to go!

RichiCoder1:
Sorry for the delay on that, I thought I had written docs. Added!

PR #450, opened 2019-01-22
Upgrades Enquirer to 2.x

jamonholmgren:
This was nontrivial to do, but I have managed to preserve backwards compatibility. I added tests as well, which helped me avoid regressions.

Unfortunately, the `list` type was switched to something completely new (comma-separated values, I guess?) so I made the choice to translate `list` to `select`. This means anyone trying to use the new `list` type will instead get the `select`. Since there are many Gluegun-powered CLIs in the wild using `list` as a `select` now, I'll just assume they want a `select`.

If someone wants the comma-separated list, they'll have to use Enquirer directly. I've added a warning about that.


GantMan:
WOAH!  epic effort!  I'll see if I can go over it with a fine-tooth comb at some point

Issue #457, opened 2019-02-17
Which Language Would You Like to Use? prompt should not accept invalid input


sw-yx:
Currently running gluegun shows this:
![image](https://user-images.githubusercontent.com/6764957/52912079-7c8e8c00-3250-11e9-8f42-2d7d2b31de78.png)

and when i type "Typescript" instead of "TypeScript" it runs a JS build instead.

expected behavior: it should allow us to pick from a limited set of options instead of relying on user entry.

i looked around the issues and couldnt believe nobody's raised this before. seems a simple dx thing to fix?

jamonholmgren:
This is strange, it's supposed to display a language picker here! Marked as a bug so we can look into this.

rdewolff:
This does not work at all for me.

rdewolff:
added an issue to track the global error : https://github.com/infinitered/gluegun/issues/462

jamonholmgren:
Fixed in #465.

PR #458, opened 2019-02-17
[minor improvement to `new` command] Print Language used and Make case insensitive

sw-yx:
slight improvement to make the Which Language prompt case insensitive (experienced here https://github.com/infinitered/gluegun/issues/457) and also show the selected language. the real fix is to use some sort of picker and not rely on user string input.

modify as you wish!


jamonholmgren:
@sw-yx Apologies for the delay on this, I've been focusing on some other projects. Looks good to me!

PR #461, opened 2019-02-25
fix links to docs

calendee:
The docs home page has links to "toolbox api docs" and "runtime docs" that result in 404s.

glue-gun-docs-links

gluegun-toolbox-bad-link

gluegun-runtime-bad-link

Update the links to point to the right location.


jamonholmgren:
Thanks @calendee !!

Issue #427, opened 2018-12-23
Displaying command specific help


mikeerickson:
Does any interface exist to display command specific help?

jamonholmgren:
@mikeerickson Hm, good question. Right now, no, but that's something we should add.

Something like:

```shell
$ movie imdb --help
```

... should display help for all commands under `./src/commands/imdb/`.

For now, you can check it in your root command:

```typescript
import { GluegunToolbox } from 'gluegun'

const HELP_MESSAGE = `
movie api
      
  Run \`movie api reset\` to reset your IMDB token.
      
`

module.exports = {
  name: 'api',
  run: async (toolbox: GluegunToolbox) => {
    const {
      meta: { commandInfo },
      print,
      parameters: { options }
    } = toolbox

    if (options.help || options.h) {
      print.info(HELP_MESSAGE)
      return
    }

    // do other things
  }
}
```

There is already a built-in capability to do this here:

https://github.com/infinitered/gluegun/blob/663972d05a01cb83852a39486a6f5f97abf7ad5b/src/toolbox/meta-tools.ts#L51-L66

However, it's not exposed with the full interface:

https://github.com/infinitered/gluegun/blob/663972d05a01cb83852a39486a6f5f97abf7ad5b/src/core-extensions/meta-extension.ts#L17

If we updated `toolbox.meta.commandInfo()` to take an argument and pass that on as the second argument, then it would be easy to implement this:

```
    const { meta, print } = toolbox
    print.info(meta.commandInfo(['api']))
```

We should also make it so this is essentially what happens when you do `mycli foo bar --help` by default, especially if you added the `.help()` call to the CLI builder.

I definitely want to do this.

RichiCoder1:
Being able to print what arguments a command expects would be nice too. Should I make that a separate ticket?

mikeerickson:
I think these two Items go hand in hand, definitely something that should be supported together. Same ticket, or separate but both should be supported.

mikeerickson:
@RichiCoder1 @jamonholmgren 

Back from holiday, figured I would chime in some more on this topic.  

I have created CLIs internally using some homegrown libraries and one of the important features is the ability to show command specific help, including support for command attributes

This is how command specific help appears:

![image](https://user-images.githubusercontent.com/183153/50607861-967e3600-0e7f-11e9-9741-746bd72e36ff.png)

The content for command help is read from the command file

```js

module.exports = {
  init: function(cli) {
[...]

PR #464, opened 2019-03-06
Add support for command-specific yargs-parser option specs

Jannis:
This allows any command to specify how its options are parsed, based on the [opts argument for yargs-parser](https://github.com/yargs/yargs-parser#requireyargs-parserargs-opts):

```typescript
module.exports = {
  description: '...',
  options: {
    boolean: ['with-x'],
  },
  run: async toolbox => ...
}
```

Without this, invokations like

```sh
mycli cmd --with-x arg1 arg2
```

would result in `arg1` being consumed as the value for `--with-x`, and `toolbox.parameters.first` being set to `arg2`. To me this is undesirable, because it really complicates handling input arguments in any non-trivial command.

The new `options` feature allows commands to control how their arguments are parsed, making it possible for

```sh
mycli cmd --with-x arg1 arg2
```

to parse to `{ withX: true, 'with-x': true }` options, with `toolbox.parameters.first` being `arg1` and `toolbox.parameters.second` being `arg2`.

The hard part of this change is detecting commands correctly without having parsed the options first. This PR is a draft because the way I implemented this is maybe not elegant and robust enough. Basically, it scans _all_ arguments (including options) and builds two separate arrays, one for the command path, the other for all the unrelated arguments. When encountering a `-foo` or `--foo` argument, it pushes it to the unrelated arguments array. The next value is then treated as a command path segment only if `[...prefix, nextValue]` is a valid command path. It is treated as an option value when encountering situations like `--option cmd cmd` however.

@skellock @jamonholmgren Curious to hear what you think!

### DIFF
--- a/src/domain/command.ts
+++ b/src/domain/command.ts
@@ -20,6 +20,8 @@ export interface GluegunCommand<TContext extends Toolbox = Toolbox> {
   file?: string
   /** A reference to the plugin that contains this command. */
   plugin?: Plugin
+  /** Hints for parsing options with yargs-parser */
+  options?: object
 }
 
 /**
@@ -35,6 +37,7 @@ export class Command implements GluegunCommand<Toolbox> {
   public alias
   public dashed
   public plugin
+  public options
 
   constructor(props?: GluegunCommand) {
     this.name = null
@@ -46,6 +49,7 @@ export class Command implements GluegunCommand<Toolbox> {
     this.alias = []
     this.dashed = false
     this.plugin = null
+    this.options = null
     if (props) Object.assign(this, props)
   }
 
--- a/src/loaders/command-loader.test.ts
+++ b/src/loaders/command-loader.test.ts
@@ -36,6 +36,9 @@ test('load command from preload', async () => {
     alias: ['z'],
     dashed: true,
     run: toolbox => 'ran!',
+    options: {
+      alias: { foo: 'f' },
+    },
   })
 
   expect(command.name).toBe('hello')
@@ -46,4 +49,5 @@ test('load command from preload', async () => {
   expect(command.file).toBe(null)
   expect(command.dashed).toBe(true)
   expect(command.commandPath).toEqual(['hello'])
+  expect(command.options).toEqual({ alias: { foo: 'f' } })
 })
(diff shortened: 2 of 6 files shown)


jamonholmgren:
I like the idea in principle! I would like to see if it causes problems with nontrivial code bases like AWS Amplify CLI, Solidarity, and the upcoming Ignite CLI.

Jannis:
Agreed. The tests are non-exhaustive and I've only tried this with https://github.com/graphprotocol/graph-cli. My feeling is that unless there is a global CLI spec built from all commands and their options, there will always be a level of ambiguity.

---

Since I didn't want to fork gluegun just for this, I've implemented a simple heuristic in an upcoming feature for graph-cli to deal with the original problem:
```js
let { withX } = toolbox.parameters.options
let array = fixParameters(toolbox.parameters, { booleanOptions: { withX } })
```
where `fixParameters` simply loops over all boolean props and builds a new parameter array from `toolbox.parameters.array`. If one of the boolean options has a string value, that value is pushed to the front of the parameters array. It's an OK workaround for now.

---

Still, if a command like `mycli cmd --debug arg1 arg2` doesn't have `arg1` and `arg2` as `first` and `second`, that can result in really weird behavior and a requires CLI developers to come up with similar workarounds. And boolean flags like `--verbose`, `--debug` are pretty common.

Potential solutions / improvements:

* Restrict options to be used at the end. Rather unusual, wouldn't recommend.

* Restrict options to be used after the commands (that way supporting `command.options` would be straight forward). Basically:
  ```sh
  mycli [cmd ...] [options] [arg ...]
  ```
  but also allowing options and args to be swapped. First, parse `[cmd ...]` to identify the command, then parse the arguments based on the yargs-parser opts spec. Could break existing scripts that put options before commands. This would be my preferred solution.

* Build a global CLI arguments spec from all commands and their options to avoid ambiguity altogether. To avoid ambiguity, which options take values and which don't needs to be known upfront. Hard.

jamonholmgren:
After thinking this over, I agree that doing it the way you explain in your original PR sounds the best. I'd love if you could bring this to a point where you feel it's testable, and then I'll go test it with Ignite and Solidarity. @Jannis thanks for your work!

danawoodman:
Is there a way to get something like this supported where we can define option types and have them auto-documented?

jamonholmgren:
@danawoodman Left a response on the other PR: https://github.com/infinitered/gluegun/issues/427#issuecomment-534706864

PR #484, opened 2019-03-12
Add ability to express after, before, delete and replace as regex

eduplus:
With these changes one can use regex for

- before
- after
- delete
- replace

values, giving them extra utility when patching a file.

So for example, can do: 

``` 
  const updated = await patching.patch(t.context.textFile, {
    replace: new RegExp('very amazing'),
    insert: 'patched info',
  })
```


jamonholmgren:
Excellent! Can you update [the patching documentation](https://github.com/infinitered/gluegun/blob/9a78cfdd9b4f00a0f150bdbf00dd04c842762e58/docs/toolbox-patching.md) to show a Regex example or two?

PR #514, opened 2019-05-04
utils types: Switch `any` to generics 

jamonholmgren:
Per @sw-yx's tweet, I switched (most of) the utils to generics.

https://twitter.com/swyx/status/1124498943801462789

I think the result isn't as pretty, but it's more useful.


sw-yx:
idk, looks pretty to me!

for object access, you will run into the `index: string` problem someday. posting a link to save you time: https://basarat.gitbooks.io/typescript/docs/types/index-signatures.html#declaring-an-index-signature

PR #517, opened 2019-05-09
It fixes typo in extension example

Lukyhenson:
### Description
It fixes typo in extension example, when i tried to create an extension following the example i got an error. i fix it using `function(toolbox) or toolbox =>`.

before
```js
// extensions/sayhello.js
module.exports = (toolbox) {
  const { print } = toolbox

  toolbox.sayhello = () => {
    print.info('Hello from an extension!')
  }
}
```
after
```js
// extensions/sayhello.js
module.exports = function(toolbox) {
  const { print } = toolbox

  toolbox.sayhello = () => {
    print.info('Hello from an extension!')
  }
}
```

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -59,7 +59,7 @@ Extensions are additional functionality that you can monkeypatch onto the `toolb
 
 ```js
 // extensions/sayhello.js
-module.exports = function(toolbox) {
+module.exports = (toolbox) => {
   const { print } = toolbox
 
   toolbox.sayhello = () => {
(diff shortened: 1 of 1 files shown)


docs/plugins.md
```diff
@@ -59,7 +59,7 @@ Extensions are additional functionality that you can monkeypatch onto the `toolb
 
 ```js
 // extensions/sayhello.js
-module.exports = (toolbox) {
+module.exports = function(toolbox) {
```
kevinvangelder: I think we would prefer:
```suggestion
module.exports = (toolbox) => {
```

Lukyhenson: i prefer that too :smile:

PR #518, opened 2019-05-11
Fixing a typo: s/console/print/

wataruoguchi:
Before: replace `console.debug(movie)` with ...

After: replace `print.debug(movie)` with ...


jamonholmgren:
Thanks @wataruoguchi !

Issue #515, opened 2019-05-06
New toolbox tools: toolbox.package


jamonholmgren:
In Ignite, we often have to check if `yarn` or `npm` are installed and run the proper command.

It would be great to have a single API to run either of those:

```js
const result = await toolbox.package.add('some-package', { ...options })
const result = await toolbox.package.remove('some-package', { ...options })
const result = await toolbox.package.list({ ...options })
```

This would use `yarn` if available, `npm` if not, and you could pass in `packageManager: 'npm'` to force one or the other.

Even just `add` and `remove` would be super helpful.

This would also be useful for when Yarn [releases their JavaScript API](https://github.com/yarnpkg/yarn/issues/6953) in that we could use it directly without having to rewrite shell commands.

PR #531, opened 2019-05-30
Adds meta.packageJSON, meta.checkForUpdate, Builder.checkForUpdates

jamonholmgren:
This adds three things:

### meta.packageJSON

This returns the contents of the CLI's packageJSON.

```js
const packageContents = toolbox.meta.packageJSON()
```

### async meta.checkForUpdate

This async function checks NPM for an update to the current CLI.

```js
const newVersion = await toolbox.meta.checkForUpdate()
if (newVersion) toolbox.print.info(`New version available: ${newVersion}`)
```

### Builder.checkForUpdates()

This builder method will allow you to set a random frequency for checking for updates and will output a simple log letting the user know there's a new version.

```js
const cli = build('movie')
  .src(__dirname)
  .checkForUpdates(5)
  .create()
```

This allows you to set a random frequency. 100 will check all the time, while 0 will never check (you can leave it off if so), and 5 will check 5% of the time.

Screen Shot 2019-05-29 at 5 05 21 PM

### DIFF
--- a/src/toolbox/meta-tools.ts
+++ b/src/toolbox/meta-tools.ts
@@ -1,13 +1,14 @@
 import { equals, replace } from './utils'
 import { GluegunToolbox } from '../domain/toolbox'
+import { PackageJSON } from './meta-types'
 
 /**
- * Finds the version for the currently running CLI.
+ * Finds the currently running CLI package.json
  *
- * @param toolbox Currently running toolbox.
- * @returns Version as a string.
+ * @param toolbox
+ * @returns Package.json contents as an object.
  */
-export function getVersion(toolbox: GluegunToolbox): string {
+export function getPackageJSON(toolbox: GluegunToolbox): PackageJSON {
   let directory = toolbox.meta.src
   const { filesystem } = toolbox
   if (!directory) throw new Error('getVersion: Unknown CLI version (no src folder found)')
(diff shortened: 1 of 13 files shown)


src/toolbox/meta-tools.ts
```diff
@@ -49,3 +61,13 @@ export function commandInfo(toolbox: GluegunToolbox, commandRoot?: string[]): st
       return [`${commandPath.join(' ')} ${alias}`, replace('$BRAND', toolbox.runtime.brand, command.description || '-')]
     })
 }
+
+export async function checkForUpdate(toolbox: GluegunToolbox): Promise<false | string> {
+  const { system, semver } = toolbox
+  const packageJSON = getPackageJSON(toolbox)
+  const myVersion = packageJSON.version
+  const packageName = packageJSON.name
+  const latestVersion = await system.run(`npm info ${packageName} dist-tags.latest`)
```
GantMan: only issue I see here is that you are assuming NPM.

I think this might be a problem in 2 ways:
1.  The user might not have npm installed.  We actually ran into this with people going full in on yarn
2.  The user might not publish lates on NPM, and might instead use the new GitHub style

It's worth fixing item 1 and adding a TODO for item 2 imo.

jamonholmgren: Yeah, I thought about this before I implemented. The problem is that it'll slow this command down even further. My thought is that we offer a `toolbox.package` core extension in the future that smartly switches between NPM and Yarn and does everything else. I have an issue filed for this, #515. When that gets implemented, I'll search the code-base for anything `npm`-related and replace with the abstraction. Until then, I'm happy to support NPM only for now.

PR #537, opened 2019-06-05
Boilerplate TypeScript CLIs now properly copy templates folder

### LINKED ISSUE #509: `build` script doesn't copy templates directory and files to build folder
So I have used the gluegun cli to create a new cli project and built myself an nice little cli. It works just fine on my machine in dev mode. But when I published it to npm and installed it globally on my other machine, I noticed a problem with the templates. It would give me a template not found error.

 I went back to look at what was generated in the build folder when running the `build` script and realized that it didn't copy the templates into it. I added them manually to the build folder and republished to npm. Reinstalled on my other machine and now it works.

Clearly, the built code expects to find the templates in the build folder but the `build` script doesn't seem to copy them in there. How can I fix this?

Here is the repo: https://github.com/snowfrogdev/Vts
You can install the npm package: npm i -g vts-cli

jamonholmgren:
Fixes #509. Ref https://github.com/snowfrogdev/Vts/pull/2.

### DIFF
--- a/src/cli/templates/cli/.gitignore.ejs
+++ b/src/cli/templates/cli/.gitignore.ejs
@@ -5,3 +5,4 @@ coverage
 .nyc_output
 dist
 build
+.vscode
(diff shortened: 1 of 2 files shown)


nirre7:
Hi, would be great to have this merged and merged into ignite-cli 👍

Issue #393, opened 2018-11-29
Better typescript support for dependencies


wgottschalk:
I'm going through the tutorial online and I'm not sure if this is a typescript thing or a gluegun thing. What I would like is to be able to define an extension:

```typescript
// src/extensions/imdb-extension.ts

import { prompt } from 'gluegun/prompt'
import * as imdb from 'imdb-api'
import { GluegunToolbox } from 'gluegun'

export interface ImdbExtension {
  imdb: {
    getMovie: (movieName: string) => Promise<imdb.Movie>
  }
}

export default (toolbox: GluegunToolbox) => {
  // memoize the API key once we retrieve it
  // let imdbKey: string | false = false

  // get a movie
  async function getMovie(movieName: string): Promise<imdb.Movie> {
    const result = await prompt.ask({ type: 'input', name: 'key', message: 'API Key>' })
    if (!result.key) return

    return imdb.get({ name: movieName }, { apiKey: result.key, timeout: 30000 })
  }

  // attach our tools to the toolbox
  toolbox.imdb = { getMovie }
}
```

AND I would like this object to show up in my typescript autocompletions. The closest thing I could think of was creating the interface above and doing some gymnastics but I feel that the typesystem could support this natively somehow.

```typescript
import { GluegunCommand, GluegunToolbox } from 'gluegun'
import { ImdbExtension } from '../extensions/imdb-extension'

const movieCommand: GluegunCommand = {
  name: 'movie',
  run: async (toolbox: GluegunToolbox) => {
    // ** Fails - Type 'GluegunToolbox' is not assignable to type 'ImdbExtension'.   Property 'imdb' is missing in type 'GluegunToolbox'. [2322]
    // const { imdb }: ImdbExtension = toolbox 

    // works
    const { imdb }: ImdbExtension = toolbox as any
    const movie = await imdb.getMovie('The Room')
    if (movie.name) toolbox.print.info(movie.name)
  }
}

export default movieCommand
```

What would be ideal
```typescript
const movie = toolbox.imdb.getMovie("The Room")   <- just autocompletes because TS knows the type
if (movie.name) toolbox.print.info(movie.name)
```

jamonholmgren:
@wgottschalk Thanks for the issue, and apologies for how long it's taken me to respond. Here's an example that I think might help you understand how to structure your type:

https://github.com/infinitered/solidarity/blob/master/src/types.ts#L43

@GantMan might have additional thoughts.

GantMan:
[...]

PR #564, opened 2019-07-27
fix(#393): generic was unused set ask response as default

ductiletoaster:
While creating a prompt to generate react components for a workflow tool. I ran into an issue where I couldn't easily tell the prompt what my expected return object would look like.

My work around was to use:
```
as any
```
However I realized that the generic T was not being refrenced so I set AskResponse as the default and made the Promise generic.

**Before**
```
const options = await prompt.ask(questions);
```
Options would be considered to be GluegunAskResponse which of course is expected
```
const options = await prompt.ask<MyResponse>(questions);
```
In this case we would still be given GluegunAskResponse which is NOT expected.

**After**
```
const options = await prompt.ask(questions);
```
Now options will default to GluegunAskResponse.
```
const options = await prompt.ask<MyResponse>(questions);
```
And this time be the correct type.

### DIFF
--- a/src/toolbox/prompt-types.ts
+++ b/src/toolbox/prompt-types.ts
@@ -4,16 +4,20 @@ const Enquirer = require('enquirer')
 export type GluegunEnquirer = typeof Enquirer
 export const GluegunEnquirer = Enquirer
 
+export interface GluegunAskResponse {
+  [key: string]: string
+}
+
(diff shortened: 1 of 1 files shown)


jamonholmgren:
This looks great! Thanks for submitting it, @ductiletoaster!

PR #577, opened 2019-08-15
fix(toolbox): `run` command potential fall through when rejecting

jperasmus:
* Promise `resolve` will no longer be called when the `system.run()`
method rejects
* Strengthened types for `system.run` slightly
* Added basic tests for `system.run`


jamonholmgren:
GREAT! 💜

PR #578, opened 2019-08-15
perf: improve performance by removing unnecessary filter loops

jperasmus:


jamonholmgren:
Looks awesome @jperasmus ! I'll merge this once it passes.

PR #579, opened 2019-08-15
chore(build): use `yarn.lock` for cache key and avoid dependency version creep

jperasmus:
- Use the `yarn.lock` lockfile's checksum instead of `package.json` because
it is more accurate since the dependencies in `package.json` are not set
to exact versions
- Additionally, also ensure that the lockfile does not update on the CI server


jamonholmgren:
Thanks so much @jperasmus ! Appreciate all your help and your patience as I get to these.

PR #587, opened 2019-08-31
fix(cli): templates can now have includes (#570)

### LINKED ISSUE #570: ejs include doesn't work
Been playing around with the package, but ejs `include` doesn't seem to work. 

Error:

```
Could not find the include file "myFile"
    at getIncludePath (app/node_modules/ejs/lib/ejs.js:165:13)
```

jonatanlins:
When an include was made inside an .ejs template, an error would occur.

The fix for the problem was to pass a filename property with the template path to the ejs.render().

Close #570.

### DIFF
--- a/src/cli/templates/test/helloMessage.ejs
+++ b/src/cli/templates/test/helloMessage.ejs
@@ -1,0 +1,1 @@
+'Hello. I am a chatty plugin.'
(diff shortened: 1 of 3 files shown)


jamonholmgren:
Looks great! Thanks so much @jonatanlins !

PR #596, opened 2019-09-11
`fancy` can support any type of message not just strings

danawoodman:
`console.log` can handle logging out any type of data, so no reason to limit to just strings


jamonholmgren:
Good call, thanks @danawoodman !

danawoodman:
🙇 🍻

PR #621, opened 2019-10-05
chore: Changed template generation to allow for other languages

jamonholmgren:
This is a fairly minor code cleanup / refactor.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -6,5 +6,4 @@ readme.md
 dist/
 build/
 docs/readme.md
-*.js.ejs
 *.ejs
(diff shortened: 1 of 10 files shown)


.prettierignore
```diff
@@ -6,3 +6,5 @@ readme.md
 dist/
 build/
 docs/readme.md
+*.js.ejs
```
bryanstearns: You don't need `*.js.ejs` if you've got `*.ejs` :-)

jamonholmgren: Making doubly-sure! Haha ... you're right.

Issue #611, opened 2019-09-30
Add `./src/types.ts` to generated TypeScript CLI


jamonholmgren:
Per #451, for TypeScript generated CLIs, I'd like to add a `./src/types.ts` file that you export your public-facing types from, plus these:

tsconfig.json:

```json
  "compilerOptions": {
    "declaration": true,
    "declarationDir": "build/types",
```

package.json:

```
  "types": "build/types/types.d.ts",
```

JessWallin:
Hey! Put up a PR for this. Let me know if this is what you're looking for, I'm happy to revise.

PR #623, opened 2019-10-06
add additional typescript options to cli output

JessWallin:
Addresses #611 

I'm not totally sure this is what you had in mind for this issue, but please let me know! I'm happy to keep working on this until it's the right solution.

### DIFF
--- a/src/cli/commands/new.ts
+++ b/src/cli/commands/new.ts
@@ -113,6 +113,7 @@ const NewCommand: GluegunCommand = {
     ]
 
     if (props.language === 'typescript') {
+      files.push('./src/types.js.ejs')
       files.push('tsconfig.json.ejs')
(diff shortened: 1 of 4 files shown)


jamonholmgren:
Looks great Jess! Thanks so much!

jamonholmgren:
I was a little quick on the trigger, @JessWallin -- we need to make sure the new file gets copied over if it's a TypeScript project.

You can do that by adding it here (in a new PR):

https://github.com/infinitered/gluegun/blob/6ca91b65ced78fea481151db64cdbb0a8e81893c/src/cli/commands/new.ts#L117-L121

JessWallin:
Submitting those changes now, @jamonholmgren

jamonholmgren:
Ref #631 but my fault, not necessary.

Issue #362, opened 2018-02-25
Generated CLIs should have the new .exclude by default


jamonholmgren:
**This is a good first-PR for someone**

Per @GantMan , let's include a commented-out line for the new `exclude` builder method.

https://github.com/infinitered/gluegun/blob/master/src/cli/templates/cli/src/cli.js.ejs#L8-L15

(While you're in there, also change to `.brand('<%= props.name %>')` to fit the new recommendations in #358)

```js
  const cli = build('<%= props.name %>')
    .src(__dirname)
    .plugins('./node_modules', { matching: '<%= props.name %>-*', hidden: true })
    .help()           // provides default for help, h, --help, -h
    .version()        // provides default for version, v, --version, -v
    // enabled the following method if you'd like to skip loading one of these core extensions
    // this can improve performance if they're not necessary
    // .exclude(['meta', 'strings', 'print', 'filesystem', 'semver', 'system', 'prompt', 'http', 'template', 'patching'])
    .create()
```

### Why?

`gluegun new foo` creates a new CLI called `foo`. If we include the `.exclude` method, it'll give new CLI authors the tools they need to optimize their CLI startup time.

JessWallin:
Added the .exclude in the PR above. It looks like the .brand had been added there already, but I'm happy to make any other tweaks if you need.

PR #624, opened 2019-10-06
add commented-out .exclude method to cli.js.es

JessWallin:
Addresses #362 

It looks like the .brand has been updated, but the .exclude method has been added to the CLI template.


jamonholmgren:
Thanks so much @JessWallin ! Really appreciate your contributions.

PR #637, opened 2019-10-16
Add yarn build as a prePublish step to the CLI default package.json

orta:
then you don't have to remember


jamonholmgren:
Thanks @orta! Great idea.

PR #655, opened 2019-11-01
Update yargs-parser to 16.1.0

friederbluemle:


jamonholmgren:
Thank you @friederbluemle! I'll release this with the next significant change.

PR #706, opened 2020-01-02
Makes toolbox patch function able to receive multiple options

LORDBABUINO:
This allows to make multiple changes in a file at once with only one file access using `toolbox.patching.patch` function. Example:

```javascript
const updated = await toolbox.patching.patch(
  textFile,
    {
      delete: 'some words',
    },
    {
      replace: 'very amazing',
      insert: 'patched info',
    }
)
```


jamonholmgren:
Thanks for your patience, @LORDBABUINO , I've been focusing on some other things lately. This patch looks amazing and I think it's very well done. I'd merge it now, but we should probably update the docs first. Would you mind doing that?

LORDBABUINO:
Hi, I updated the docs and fixed some bugs on patch. The function was crashing when it receives an `insert` that is already present in the file.

jamonholmgren:
@LORDBABUINO Thank you so much! Released now. :)

PR #714, opened 2020-02-28
doc: add an example about the use of props in a template

EmmanuelDemey:


AdrielBento:
Simplify the example, it looks much better.

jamonholmgren:
Thanks for helping out on Gluegun, @EmmanuelDemey ! 💜

Issue #573, opened 2019-08-06
loadConfig Documentation Parameters Reversed


blkmutt:
In the documentation (https://github.com/infinitered/gluegun/blob/master/docs/toolbox-config.md), parameters appear to be reversed for loadConfig. The samples appear to be as follows: 

`const myConfig = loadConfig(process.cwd(), brand)`

Looking at src/loaders/config-loader.ts, it appears as though the function call is looking for name and then directory. 

`/**
 * Loads the config for the app via CosmicConfig by searching in a few places.
 *
 * @param name The base name of the config to load.
 * @param src The directory to look in.
 */
export function loadConfig(name: string, src: string)`

I could update either the documentation or the code and submit a PR, but would you prefer the documentation remains the same or the code?

jamonholmgren:
Hey @blkmutt let's update the documentation -- thank you so much for catching this! I'd probably prefer how it's documented but we've already had it out in the wild with the other order for a while, so we'd better keep it to reduce annoyance/bugs on upgrading.

PR #722, opened 2020-07-07
fix(template): match documentation order

HurricaneInteractive:
Update `cli-extension` template to match the parameter order of the [documentation](https://infinitered.github.io/gluegun/#/toolbox-config). The order was also discussed in #573.


jamonholmgren:
Thank you!

PR #727, opened 2020-10-18
run command -- adds stdout to error object when it throws an error

jamonholmgren:
Need this for Ignite Flame ... have to analyze the stdout on an error thrown for tests.


jamonholmgren:
Windows CI is failing, but I need to push out a new version for Ignite. We'll fix the Windows CI later.

PR #729, opened 2021-01-13
Fix typo in toolbox-parameters-docs

kathawala:
I think the following should be true

```js
toolbox.parameters.string.split(' ') == toolbox.parameters.array
```

If that's correct, then the docs have a typo where `string` and `array` options of `toolbox.parameters` don't have their examples matching up with each other


jamonholmgren (approved):
Good catch!

PR #731, opened 2021-01-16
Add new methods to toolbox.print (highlight and muted) (Fixes #726)

### LINKED ISSUE #726: New colors theme method and improved colors theme docs
Hey, a couple of quick suggestions I'm happy to submit a PR for if you're interested:

1. Add a new colors theme method `command` which points to `magenta`

I like to use a distinct and consistent color when telling the user they can use another command to help with something (e.g. from my CLI `` `This action requires a logged in user. Run ${magenta(`${brand} login`)} to log in.` ``). Currently the default command prints `` `Type ${magenta(`${brand} --help`)} to view common commands.` `` so I think `magenta` is a good candidate for this.

2. Document all the theme methods in the docs

Currently [the docs](https://infinitered.github.io/gluegun/#/toolbox-print?id=colors) don't document the following theme methods:

* `highlight`
* `line`

I especially wanted to use the same consistent highlight color as the default commands, and had been using `cyan` until I was snooping through the source and discovered the `highlight` method. I think it would be useful to expose all of them.

kathawala:
A few of the colors in `toolbox.print.colors` are left out of the docs and also don't have any print function associated to them:

* `highlight`
* `muted`

This PR adds documentation about those colors and adds functions for them so one can do

```js
toolbox.print.highlight('important text')
```

the same way it's currently possible to do

```js
toolbox.print.success('important text')
```

This was asked for in #726 and I myself wrote a quick utility to help print those functions, so I thought it might be worth adding those function here directly.


jamonholmgren (approved):
This looks great! I don't think the test failure is your problem.

Issue #724, opened 2020-08-24
support for passing style option to print.table


Faizahmadfaiz:
Hey guys, I want to print a table in which I want some of my column cells to be right aligned. Going through **cli-table3** (which is used by gluegun) examples, I found this 
```
var table = new Table({
  chars: { 'top': '' , 'top-mid': '' , 'top-left': '' , 'top-right': ''
         , 'bottom': '' , 'bottom-mid': '' , 'bottom-left': '' , 'bottom-right': ''
         , 'left': '' , 'left-mid': '' , 'mid': '' , 'mid-mid': ''
         , 'right': '' , 'right-mid': '' , 'middle': ' ' },
  style: { 'padding-left': 0, 'padding-right': 0 }
});
```
Currently in table function I can see we can pass only **data: string[][], options: GluegunPrintTableOptions = {}** .It seems to me currently I can't achive my goal of making cells right aligned.

jamonholmgren:
@Faizahmadfaiz I'd accept a PR that adds it to `GluegunPrintTableOptions`!

PR #736, opened 2021-05-14
feat(print): add style options for tables

PauloRSF:
This adds the ability to pass a `style` property on `print.table`'s options, as requested on #724 (although it still doesn't fully achieve what the issue's author wanted).

* Created new type `TableStyle`
* Added style prop on `GluegunPrintTableOptions`
* Modified markdown table formatting to correctly print columns header dividers when padding is applied
* Updated tests and docs accordingly

### DIFF
--- a/docs/toolbox-print.md
+++ b/docs/toolbox-print.md
@@ -193,3 +193,38 @@ Output:
 | Steve      | Kellock   | 43  |
 | Gary       | Busey     | 73  |
 ```
+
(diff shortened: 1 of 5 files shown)


jamonholmgren (approved):
Nice! This is great. Thank you!

jamonholmgren:
Apologies for not seeing this. I think it got lost in my notifications.

PR #750, opened 2022-01-24
Clarify Amplify's use of Gluegun - fixes #747

### LINKED ISSUE #747: Remove aws-amplify from the 'Who Is Using This?' section
I'm checking the aws-amplify repository to see how it apply gluegun but seem like all reference to gluegun is already removed in [aws](https://github.com/aws-amplify/amplify-cli/pull/2254/files) (except for some changelog and comment)

I think we should remove it to prevent confusion for next developer want to check for reference.

jamonholmgren:
Amplify used to use Gluegun for the first versions of their CLI but has since integrated Gluegun's functionality into their CLI in a custom way. This clarifies the wording around that.

Thanks to @chigia001 for the issue.


derekgreenberg (approved):
This text change to the README appears to be correct to me.

Issue #753, opened 2022-02-09
toolbox.prompt overwrites all items


imranMnts:
**Describe the bug**
`toolbox.prompt.ask` overwrites all items

**To Reproduce**

Issue :
```
let addNewItem = true;
const result = [];
while (addNewItem) {
  const item = await toolbox.prompt.ask({
      type: 'input',
      name: `name`,
      message: `message`,
  });

  result.push(item);
  addNewItem = await toolbox.prompt.confirm('continue?');
}
console.log('result: ', result);
```

the first time, we put `1` as name input, and `2` the second time

**Result**

print: 
```result:  [ { name: '2', yesno: false }, { name: '2', yesno: false } ]```

**Expected behavior**
print: 
```result:  [ { name: '1' }, { name: '2' } ]```

**Doctor (please complete the following information):**
 - OS: 11.6.2
 - Gluegun Version: 5.1.0

**Comment by jamonholmgren** [comment]
@imranMnts Thanks for the bug report. Would you mind taking a shot at a PR?

**Comment by jamonholmgren** [comment]
This would be a good opportunity for a regression test as well.

**Comment by imranMnts** [comment]
@jamonholmgren I created this PR https://github.com/infinitered/gluegun/pull/754

can you please have a look?

**Comment by imranMnts** [comment]
@jamonholmgren 

Can we have an update, please?

**Comment by ugur-eren** [comment]
I guess you can also use destructuring assignment to get the name as seen on the [docs](https://infinitered.github.io/gluegun/#/toolbox-prompt?id=ask), while waiting for Jamon's thoughts about this. So you would avoid passing the object reference to the array.

```javascript
const {name} = await toolbox.prompt.ask({
    type: 'input',
    name: `name`,
    message: `message`,
});

result.push({name});
```
Or, if you're using `ask` with an array to ask many questions at once, you can use spread syntax to get all responses.

```javascript
const {...item} = await toolbox.prompt.ask({
    type: 'input',
    name: `name`,
    message: `message`,
});

result.push({...item});
```

PR #754, opened 2022-02-10
fix ovewrite issue #753

imranMnts:
Hello @jamonholmgren 

Don't know the reason to reuse the same `enquirer` instance but when I use a new one for each `prompt` command, the issue disappears on my side.

I took the initiative to add a .npmignore and remove documentation from the npm package.
Just a suggestion, I can remove it if you need it for any reason.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,5 +1,4 @@
 tsconfig.json
 docs/
 
-LICENSE
 readme.md
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,5 @@
+tsconfig.json
+docs/
+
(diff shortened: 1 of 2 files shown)


.npmignore
```diff
@@ -0,0 +1,5 @@
+tsconfig.json
+docs/
+
+LICENSE
```
jamonholmgren, line 4:
```suggestion
```

Let's keep the license in the package.

jamonholmgren (approved):
Sorry it's taken me so long to see this! I think this is reasonable. Let's do it!

PR #755, opened 2022-02-24
fix(docs): remove ./docs from links in docs/README.md

akrabdev:
Fix getting page not found, when using links in docs/README.md

The issue:

docs-readme-issue

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -103,9 +103,9 @@ module.exports = {
 }
 ```
 
-See the [toolbox api docs](./docs/toolbox-api.md) for more details on what you can do.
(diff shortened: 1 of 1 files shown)


jamonholmgren:
Honestly, having two readmes is getting problematic, as copying one over to the other to keep them in sync results in issues like this. Anyone have suggestions? The only real reason the readme is in ./docs is to keep this website up to date, which I rarely look at: https://infinitered.github.io/gluegun

Perhaps we should remove that website and just refer people to the docs in the repo itself.

akrabdev:
When it comes to docs some prefer a dedicated site over Github markdown.

jamonholmgren:
We're working on a better solution -- thank you though, @akrabdev !

PR #763, opened 2022-08-25
Coalesce boolean arguments to actual Boolean type instead of string

joshuayoes:
When we pass `--key=false` flag arguments to gluegun, we get `{ options: { key: 'false' } }` instead of `{ options: { key: 'false' } }`. 

This isn't how I expect to consume these in commands. This can handled on the program level, but I feel like that it should be handled on the gluegun level since I can't imagine a use case where we'd want true or false to be a string

### DIFF
--- a/src/toolbox/parameter-tools.ts
+++ b/src/toolbox/parameter-tools.ts
@@ -1,6 +1,7 @@
 import { GluegunParameters } from '../domain/toolbox'
 import { Options } from '../domain/options'
 import { equals, is } from './utils'
+import type yargsParser from 'yargs-parser'
 
 const COMMAND_DELIMITER = ' '
 
@@ -12,7 +13,7 @@ const COMMAND_DELIMITER = ' '
  * @returns Normalized parameters.
  */
 export function parseParams(commandArray: string | string[], extraOpts: Options = {}): GluegunParameters {
-  const yargsParse = require('yargs-parser')
+  const yargsParse: yargsParser.Parser = require('yargs-parser')
 
   // use the command line args if not passed in
   if (is(String, commandArray)) {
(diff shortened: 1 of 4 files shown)


joshuayoes:
I've added failing tests to demonstrate desired behavior

joshuayoes:
It also may be better to take a [PR like this](https://github.com/infinitered/gluegun/pull/464), where we could provide the `yargs-parser` object optionally, and then allow for opting into coalescing options

joshuayoes:
Got some input from Jamon. Gluegun already doesn't have great type-safety guarantees on command parsing, so this quality of life improvement would be worth a potentially breaking change

joshuayoes:
Screen Shot 2022-08-25 at 4 43 04 PM
https://dev.azure.com/infinitered/gluegun/_build?view=runs

I think something changed about ts-node or typescript dependency between now and May. 

I don't think these changes are causing this issue, but I will investigate more before merging

src/toolbox/parameter-tools.ts
```diff
@@ -1,6 +1,7 @@
 import { GluegunParameters } from '../domain/toolbox'
 import { Options } from '../domain/options'
 import { equals, is } from './utils'
+import type yargsParser from 'yargs-parser'
```
jamonholmgren, line 4: Smart to use an `import type` here.

src/toolbox/parameter-tools.ts
```diff
@@ -31,7 +32,19 @@ export function parseParams(commandArray: string | string[], extraOpts: Options
...
+    Object.entries(parsed).map(([key, value]) => {
+      // if value is 'true' or 'false', convert to boolean
+      if (value === 'true') {
+        return [key, true]
+      }
+      if (value === 'false') {
+        return [key, false]
+      }
+      return [key, value]
+    }),
```
jamonholmgren, line 45: I'd use the simplified version here that I mentioned in the related Ignite PR.

jamonholmgren (changes requested):
Good to go (after CI fixed)

I'll merge & release once CI is fixed and you slim down that helper function.

jamonholmgren:
Merged into #783 since this will be a breaking change too.

PR #769, opened 2022-10-20
Use --unsafe-perm flag to resolve npm install error in circleci

silasjmatson:
The circleci setup should really be updated as it's outdated, but this works as a stopgap to resolve the current issues.


jamonholmgren:
Closing in favor of #783.

PR #776, opened 2023-04-08
update typescript to version 5.0.4

gilsongabriel:
Update typescript version to resolve this error when using typescript mode.

Error  ts.Debug.assert(typeof typeReferenceDirectiveName === "string", "Non-string value passed to `ts.resolveTypeReferenceDirective`, likely by a wrapping package working with an outdated `resolveTypeReferenceDirectives` signature. This is probably not a problem in TS itself.");

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,7 +115,7 @@
     "temp-write": "4.0.0",
     "ts-jest": "^27.1.3",
     "ts-node": "10.4.0",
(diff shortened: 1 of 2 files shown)


jamonholmgren:
Closing in favor of #783. But thank you @gilsongabriel! Hope you continue to contribute!

Issue #774, opened 2023-02-04
Critial vulnerability at ejs


carolinevsboliveira:
ejs 3.1.6 has a template injection vulnerability: https://nvd.nist.gov/vuln/detail/CVE-2022-29078. This is fixed in 3.1.7 and later.

PR #778, opened 2023-08-11
fix: ejs critial vulnerability

0xNawara:
https://github.com/infinitered/gluegun/issues/774


jamonholmgren:
Not worried about the Azure failure for now. Thanks!

PR #781, opened 2023-08-14
Use current LTS node container (18.17.1) for circleCI builds

trevor-coleman:
## 1. Updated version of node container

Previously we were using `circleci/node:14.17` but that is no longer supported. I've updated it to use `cimg/node:18.17.1` which has the current LTS version.

We could consider using the `cimg/lts` alias which would auto update to the latest LTS release, but that could introduce some instability. Probably better to just suffer through updating it once every 2-3 years when the current version ages out.

[More info here](https://circleci.com/developer/images/image/cimg/node)

## 2. Pinned versions of typescript and set sourceMap:true in the template files 

The test that was failing was a test of the generated template. It was failing the resolved version of typescript is incompatible with the version of `@typescript-eslint/typescript-estree` required by `@typescript-eslint/eslint-plugin`

`@typescript-eslint/typescript-estree` doesn't support typescript >4.5.0, so I changed the dependency from `^4.2.3` to `~4.5.0`.

Will look into whether upgrading is worth the effort. I think we'd be jumping several major versions on the eslint side.

## 3. Added `.idea` to the gitignore file

This ignores local project files auto-created by intellij IDEs when the project is opened. Just a convenience for people who use intellij IDEs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@
 defaults: &defaults
   docker:
     # Choose the version of Node you want here
-    - image: circleci/node:14.17
+    - image: cimg/node:18.17.1
   working_directory: ~/repo
 
 version: 2
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,6 @@ http.js
 patching.js
 prompt.js
 package-manager.js
+
+# Webstorm & IntelliJ
+.idea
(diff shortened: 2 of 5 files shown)


trevor-coleman:
Good to go!

jamonholmgren:
@trevor-coleman Do we need to cut a release? Looks like just a `chore` release maybe?

jamonholmgren (approved):

trevor-coleman:
> @trevor-coleman Do we need to cut a release? Looks like just a chore release maybe?

Yeah I think so -- the template will be broken for new projects until this is fixed.

trevor-coleman:
Basically the template installs a version of the template that is incompatible with the ts parser that's bundled.

jamonholmgren:
Oops -- don't merge like that @trevor-coleman

PR #783, opened 2023-09-16
Switching to Bun for dev

jamonholmgren:
# Switches to Bun

CleanShot 2023-09-15 at 18 24 20@2x

This PR switches to using [Bun](https://bun.sh) as the development runtime. This is an experiment; however, it's way, WAY faster to run. Like 20x+ faster in my tests.

**Gluegun will continue to work in Node-powered CLIs** (and we need to do extensive testing to ensure that is the case). Over time, the gap between the two runtimes will converge. And, if it works well enough, we will recommend that others use Bun instead of Node.

Running this with Bun would mean doing something like this:

```
bunx --bun gluegun ...
```

Jarred said that they're looking to make this better. https://twitter.com/jarredsumner/status/1702840231584743485

You can also run `bunx gluegun-bun ...` (theoretically), and locally you can link it with `bun link` and then `gluegun-bun ...` to force using bun.

## Removes the ability to generate JS (non-TypeScript) CLIs using `gluegun new`

This has been a long time coming. The reasons for removing the JS-only template are that we do not use Gluegun to build CLIs without TypeScript anymore and this also simplifies the template quite a bit.

Yes, I could make this a separate PR. Why do you ask? 😅


jamonholmgren:
Oh, and thanks to @cmgriffing for the bun orb!

PR #784, opened 2023-10-10
use associated `console.` functions when printing

passabilities:
* when printing an **error** or **warning** it should be sent to `stderr`
* uses associated underlying `console.` functions when printing


jamonholmgren:
I agree we should do this.

It will be a breaking change, so I'll keep it for the next version of Gluegun!

PR #785, opened 2023-10-11
Convert docs to ir-docs

trevor-coleman:
Reorganize and convert documentation to work with new unified IR docusaurus site:

* Creates ir-docs folder
* Organizes docs in folders matching current sidebar
* Adds CI scripts 
    * pnly manually-triggered script is active
    * commented out others to prevent build failures from blocking other PRs if this gets merged.
* Creates an `./ir-docs/` folder with the new documentation, this will allow the existing documentation to remain up while we make the transition. When the new docs are ready we can simply rename the folder and point the CI at it.
* Updates internal links to be relative
* Creates metadata files for categories etc.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -75,16 +75,16 @@ jobs:
 
 # Parameters for the publish-docs orb
 publish-details: &publish-details
-  description: "Use MLKit Models in React Native apps"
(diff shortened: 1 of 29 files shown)


jamonholmgren:
I'm okay deleting the existing docs website right now.

PR #795, opened 2025-01-30
CI: Bump to node 20 / LTS version

jamonholmgren:
Bumps node version to v20 / LTS to keep pace with LTS.


jamonholmgren:
Ran out of time, will revisit later

PR #796, opened 2025-03-09
Upgrade dependencies to solve audits

boieva-maryna:
Before
audit before

After
audit after


Raiper34:
Solved in https://github.com/infinitered/gluegun/pull/798
Can we close it?

